PR #1, opened 2016-07-20
Add build toolchain for browser extensions

nickstenning:
This PR makes it possible to build both Chrome and Firefox extensions using simple JSON settings files as the basis for different build configurations.

The basic approach is as follows:

- Import (copy and paste) a few utility modules from `h` which the extension depended on
- Add a `package.json` with the build dependencies of the extension
- Make the tests pass (run `npm test`)
- Move the browserified extension bundle into `src/common/`
- Convert `manifest.json.jinja2` into a mustache template
- Add mustache versions of `app.html` and `embed.js`
- Get tests and extension builds running on Travis

Obviously there's more to do here -- we need to add a README, for one -- but I figure we should get this merged and move on.

For now, if you want to test the built extensions locally, you'll need to ensure you're running the latest version of the client. Something like the following should do it:

```
git clone git@github.com:hypothesis/browser-extension
cd browser-extension
git checkout build
npm install
npm link hypothesis
make
```

At that point you should be able to load `build/` in Chrome as an unpacked extension. By default the extension will be built as a production Chrome extension. To build a production Firefox extension, you'd do:

```
make SETTINGS_FILE=settings/firefox-prod.json clean extension
```


src/chrome/manifest.json.mustache
```diff
@@ -2,14 +2,14 @@
...
-  {% if browser == 'chrome' %}
+  {{#browserIsChrome}}
   "minimum_chrome_version": "38",
-  {% endif %}
+  {{/browserIsChrome}}
 
-  {% if browser == 'firefox' %}
+  {{#browserIsFirefox}}
   "applications": {
     "gecko": {
```
robertknight: The Jinja comment tag in this section needs to be changed to a mustache comment tag, otherwise the result is invalid JSON. Apart from that, loading the Firefox add-in via the "Load Temporary Add-on" button at `about:debugging#addons` worked.

src/client/app.html.mustache
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <base href="/">
```
robertknight: Possibly worth adding a comment in here that the `<base>` tag is needed to support Angular's `$location` service in HTML 5 mode. That will serve as a reminder that we should be able to drop this requirement in future.

src/common/sidebar-injector.js
```diff
@@ -303,7 +303,7 @@ function SidebarInjector(chromeTabs, dependencies) {
     if (!isSupportedURL(tab.url)) {
       return Promise.resolve();
     }
-    return injectScript(tab.id, '/public/destroy.js');
+    return injectScript(tab.id, '/lib/destroy.js');
```
robertknight: `destroy.js` is actually not Chrome-specific at all, so it can go in `src/common`

src/chrome/manifest.json.mustache
nickstenning: Ooops! Good spot.

src/common/sidebar-injector.js
nickstenning: It needs to be somewhere where it can be injected into the page. Nothing in `src/common/` is publicly accessible.

tests/karma-phantomjs-polyfill.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+// Basic polyfills for APIs which are supported natively
+// by all browsers we support (IE >= 10)
+require('core-js/es5');
```
robertknight: Turns out that we don't actually need `ES5` polyfills any more, I think that's because of the Phantom 2.x upgrade.

src/chrome/manifest.json.mustache
nickstenning: Fixed.

robertknight:
I've tested this out locally and was able to build working Chrome and Firefox extensions that used local checkout of the `client` repo 👍 . Barring a couple of comments about directory naming, this is good to merge.

1. In our other repos, we name the unit test directory `test` (singular) rather than `tests`, and have the test directory alongside the source files, eg. `src/directive/test` rather than `tests/directive`. The advantage is that the imports of the module under test can be mostly of the form `var moduleToTest = require('../module-to-test')` rather than `var moduleToTest = require('../../src/module-to-test')`.
2. In our other repos, utility scripts are in a `scripts` directory rather than `tools` and non-shell scripts have a file extension. I'd like to be consistent unless there is a good reason not to be.

Finally, I do have some slight reservations about a make-only build setup, after I ran into a few snafus because I made assumptions that work for the client and `h` repos but not this one. This doesn't prevent me merging this PR but I wanted to document them anyway, because I'm sure it is going to catch someone else out.

- I deleted the build dir at one point and re-ran make, not realizing that it was part of the source tree. In our other repos, the `build` dir is automatically created if it does not exist
- I initially built the extension using the `hypothesis` package from npm, then switched to the `npm link`-ed version of the package. Re-running `make` after this did not update the client assets in the build dir.
- At one point I ran `make` with a different `SETTINGS_FILE` value but this had no effect, requiring a `make clean` or `make --always-make` first.

One of the upsides of our gulp toolchain in the other projects is that although it does do unnecessary work (though that can be addressed), the result reliably works after running `gulp build`.

nickstenning:
> In our other repos, we name the unit test directory test (singular) rather than tests

https://github.com/hypothesis/h/tree/master/tests 🤔

There are advantages and disadvantages to both approaches. Over the past year or so I'm increasingly of the opinion that keeping tests in an entirely separate part of the tree is preferable:

- it makes it really obvious where the tests are
- it simplifies builds (because you don't need to remember to exclude test directories)
- it simplifies configuring linters differently for app code and test code
- in Python-land, it *ensures* that you write tests against the installed package rather than the source on disk next to you.

In short: this was a deliberate decision to bring these tests in line with the tests in the `h` repository. I'd suggest we think about making similar changes to the client repo in due course.

> In our other repos, utility scripts are in a scripts directory rather than tools and non-shell scripts have a file extension. I'd like to be consistent unless there is a good reason not to be.

Happy to be consistent, but I'd like to converge on this pattern rather than the reverse. There's really no need for file extensions (that's what a shebang is for) and `tools` is more clearly about development tooling rather than `scripts` which could be JavaScript.

Your concerns about a `make`-based build are noted, but everything you raised is totally fixable without moving to gulp. I knocked up this Makefile in about 3 hours yesterday, so it's perhaps not surprising that it's not yet perfect ;).

nickstenning:
I've addressed your second and third concerns about the Makefile. I'm going to leave the first one for now as it is usually easier to not have to create build directories from a Makefile.

robertknight:
> https://github.com/hypothesis/h/tree/master/tests 🤔

Right, I was still looking at the front-end tests in `h` which are still organized in the previous way.

> In short: this was a deliberate decision to bring these tests in line with the tests in the h repository.
> I'd suggest we think about making similar changes to the client repo in due course.

Thanks for the explanation. So far it has been quite convenient having the tests live close to the source in the tree but I see your point.

> There's really no need for file extensions (that's what a shebang is for)

One advantage is that tools which only use very dumb (ie. filename-based) language detection can still recognize the language. eg. ESLint just looks at file extensions when deciding what to process and my editor shows different icons depending on the source file type. Neither super-critical but worth three extra chars in the filename IMO.

> tools is more clearly about development tooling rather than scripts which could be JavaScript.

Agreed

robertknight:
Re-tested Firefox and Chrome extension builds plus making changes in the client repo and then rebuilding the extension. That's all working nicely now.

PR #2, opened 2016-07-22
Automatically generate a sensible extension version number

nickstenning:
This commit sets up the build to automatically generate a sensible version number and version name on the basis of:

- the latest git tag, which is assumed to reflect the version of the underlying client
- the number of commits since that tag
- whether or not the build is a 'production' build

### DIFF
--- a/tools/settings.js
+++ b/tools/settings.js
@@ -1,0 +1,54 @@
+#!/usr/bin/env node
+
+/**
+ * Outputs a JSON object representing the extension settings based on a settings
+ * file and the package environment.
+ */
+
+'use strict';
+
+const path = require('path');
+const gitDescribeSync = require('git-describe').gitDescribeSync;
+
+// Suppress (expected) EPIPE errors on STDOUT
+process.stdout.on('error', err => {
+  if (err.code === 'EPIPE') { process.exit(); }
+});
+
+/**
+ * getVersion fetches the current version from git, applying the following
+ * rules:
+ *
+ * - If buildType is 'production' and the git state is not clean, throw an error
+ * - Set the version number to X.Y.Z.W, where X.Y.Z is the last tagged release
+ *   and W is the number of commits since that release.
+ * - If the buildType is 'production', set the version name to "Official Build",
+ *   otherwise set it to a string of the form "gXXXXXXX[.dirty]" to reflect the
+ *   exact commit and state of the repository.
+ */
+function getVersion(buildType) {
+  const gitInfo = gitDescribeSync();
+
+  if (buildType === 'production' && gitInfo.dirty) {
+    throw new Error('cannot create production build with dirty git state!');
+  }
+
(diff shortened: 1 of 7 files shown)


tools/settings.js
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Outputs a JSON object representing the extension settings based on a settings
+ * file and the package environment.
+ */
+
+'use strict';
+
+const path = require('path');
+const gitDescribeSync = require('git-describe').gitDescribeSync;
```
robertknight: Just a thought - If we were using Node 6 this could be written as `const { gitDescribeSync } = require('git-describe');`

tools/settings.js
```diff
@@ -0,0 +1,57 @@
...
+
+// Suppress (expected) EPIPE errors on STDOUT
+process.stdout.on('error', err => {
+  if (err.code === 'EPIPE') process.exit();
```
robertknight: Nit. Code style is to always use braces around if/while/for bodies.

tools/settings.js
```diff
@@ -0,0 +1,57 @@
...
+
+  const version = `${gitInfo.semver}.${gitInfo.distance}`;
+  let versionName = 'Official Build';
+
+  if (buildType !== 'production') {
+    versionName = `${gitInfo.hash}${gitInfo.dirty ? '.dirty' : ''}`;
+  }
+
+  return {
+    version: version,
```
robertknight: I think this can just be written as `return { version, versionName }` in Node v4.

tools/settings.js
```diff
@@ -0,0 +1,57 @@
...
+ * Outputs a JSON object representing the extension settings based on a settings
+ * file and the package environment.
+ */
+
+'use strict';
+
+const path = require('path');
+const gitDescribeSync = require('git-describe').gitDescribeSync;
+
+// Suppress (expected) EPIPE errors on STDOUT
```
robertknight: Can you explain what's going on here?

nickstenning: #### `test.js`:
```javascript
console.log('foo')
console.log('bar')
```
Then:
```bash
$ node test.js | head -1
foo
events.js:141
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at Object.exports._errnoException (util.js:874:11)
    at exports._exceptionWithHostPort (util.js:897:20)
    at WriteWrap.afterWrite (net.js:763:14)
```

Basically: Node decides that unlike every other major scripting language it's the programmer's job to catch the EPIPE that gets returned from the `write(2)` syscall when the output is truncated with SIGPIPE.

nickstenning: Oh and if it's not clear calling `tools/settings.js <settings_file> | cmp -s ...` in the Makefile results in this scenario.

robertknight:
LGTM

PR #3, opened 2016-07-25
Port some build and troubleshooting docs from the h repo

nickstenning:
Copies across (with some tweaks) the documentation from the old repository.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -1,0 +1,51 @@
+Building the extension
+======================
+
+This document describes how to build the browser extension and load it into your
+browser.
+
+Installing the dependencies
+---------------------------
+
+You should have [Node v4.x][node4] installed.
+
+Before continuing, install the development dependencies by running the following
+at the root of the repository:
+
+    $ npm install
+
+[node4]: https://nodejs.org/en/download/
+
+Building the extension
+----------------------
+
+The extension build is configured by a JSON settings file, some examples of
+which are supplied in the `settings/` directory. To build the extension, you
+simply run `make`. To build the extension in its default configuration (a Chrome
+development build), run:
+
+    $ make
+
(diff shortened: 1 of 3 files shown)


docs/troubleshooting.md
```diff
@@ -0,0 +1,34 @@
+Troubleshooting
+---------------
+
+Here are some errors you might encounter while developing the extension, and
```
robertknight: This is fine, although I'll note that these are all conditions that could be checked quite easily in the code that handles the injection. So we could potentially display a useful error page to developers (via the existing help page system) rather than having errors hidden away in the console.

docs/building.md
```diff
@@ -0,0 +1,50 @@
...
+development build), run:
+
+    $ make
+
+To build the extension from an alternate settings file, simply specify the
+`SETTINGS_FILE` variable when running `make`:
+
+    $ make SETTINGS_FILE=settings/chrome-prod.json
+
+This will build a production extension.
```
robertknight: I think we should be a little clearer here about what a "production extension" means, since it is the easiest way to build a customized version of the client.  Perhaps: "This will build a Chrome extension that uses the main Hypothesis web service."

docs/troubleshooting.md
nickstenning: Ok, good to know.

docs/building.md
nickstenning: Added a line to that effect.

docs/troubleshooting.md
nickstenning: FWIW I didn't write these -- I was just porting the docs so we don't lose them.

robertknight:
LGTM. There are some areas that would be good to expand on in future, such as how to use the Firefox extension.

PR #5, opened 2016-08-01
Setup ESLint using shared config

robertknight:
This sets up ESLint using the same shared config as the `client` and `h` repos.

There are three commits in this PR:

- The first adds the ESLint config and a `make lint` task
- The second fixes up trivial issues using `eslint . --fix`
- The third includes manual fixes for the remaining issues, which included several missing function returns, missing or misplaced 'use strict' and parenthesis in the wrong place in a function call

### DIFF
--- a/src/chrome/lib/destroy.js
+++ b/src/chrome/lib/destroy.js
@@ -1,15 +1,14 @@
 // Script injected into the page to trigger removal of any existing instances
 // of the Hypothesis client.
-(function () {
-  'use strict';
 
-  var annotatorLink =
-    document.querySelector('link[type="application/annotator+html"]');
+'use strict';
 
-  if (annotatorLink) {
-    // Dispatch a 'destroy' event which is handled by the code in
-    // annotator/main.js to remove the client.
-    var destroyEvent = new Event('destroy');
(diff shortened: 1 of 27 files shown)


src/chrome/lib/destroy.js
```diff
@@ -1,15 +1,14 @@
 // Script injected into the page to trigger removal of any existing instances
 // of the Hypothesis client.
-(function () {
```
nickstenning: Removing this IIFE means that `annotatorLink` now gets exposed on `window`, which I'm pretty sure is not what we want.

nickstenning: I'd guess you probably stop eslint from doing this automatically by passing `document` into the function:

```javascript
(function (document) {
  ...
}(window.document));
```

robertknight: This script is now run in an isolated world as a content script. That's why it destroys the annotator by sending a DOM event rather than accessing `window.annotator` directly - which it can't see.

nickstenning: Ah. Right. Ok!

nickstenning:
LGTM.

PR #6, opened 2016-09-13
Add a script that creates and pushes a new tag

robertknight:
It serves mostly as documentation for the current release process and a reference for when we come to automate the process entirely in future.

### DIFF
--- a/tools/release-it
+++ b/tools/release-it
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+# Create a new tag release of the browser extension.
+#
+# The current workflow for a browser extension release is:
+#
+#  1. Update the 'hypothesis' package dependency to the latest version
+#  2. Update the package version to match the 'hypothesis' client package
+#     version
+#  3. Tag and push a new release of the browser-extension repo using this
+#     script
+#  4. When the Travis CI builds for the tag complete, upload the
+#     extension to the Chrome Web Store
+
+set -eu
+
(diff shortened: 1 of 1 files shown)


tools/release-it
```diff
@@ -0,0 +1,26 @@
...
+#     extension to the Chrome Web Store
+
+set -e
+
+VERSION=$(node -e "console.log(require('./package.json').version)")
+TAG_NAME=v$VERSION
+
+git fetch --tags
+git tag --sign $TAG_NAME -m "Release $TAG_NAME"
+git push --tags
```
nickstenning: `git push --follow-tags` would be safer here.

tools/release-it
```diff
@@ -0,0 +1,26 @@
...
+#
+#  1. Update the 'hypothesis' package dependency to the latest version
+#  2. Update the package version to match the 'hypothesis' client package
+#     version
+#  3. Tag and push a new release of the browser-extension repo using this
+#     script
+#  4. When the Travis CI builds for the tag complete, upload the
+#     extension to the Chrome Web Store
+
+set -e
```
nickstenning: It's also good practice to `set -u` in order to warn about unset variables.

chdorner: And `set -eu` will make you feel like you're part of the EU many years from now!

robertknight: Thanks. I didn't know about `set -u`. On my system at least, `set -eu` aborts the script rather than just warning about unset variables, which is good.

PR #8, opened 2016-09-19
build env in containers using compose

cove:
See https://github.com/hypothesis/h/pull/3888


nickstenning:
As previously discussed, I don't think we currently want or need to build the browser extension in a Docker container.

PR #9, opened 2016-09-26
Update PDF.js

robertknight:
_**This depends on the client PR https://github.com/hypothesis/client/pull/134**_

This PR updates PDF.js to the current version and adds a script to document the process of updating PDF.js and make it easier to update in future.

The first commit adds a script `tools/update-pdfjs` which fetches the latest version of PDF.js from the GitHub viewer demo page, patches the viewer code to load the Hypothesis client and remove the origin check for the loaded PDF.

The second commit removes the previous version of PDF.js from `src/chrome/content` and adds the latest version in the more obviously-named location `src/vendor/pdfjs`.


sean-roberts:
From offline conversation, Robert mentioned that he missed a file push with his changes.

robertknight:
The previous commit failed to build due to a subdirectory of `src/vendor/pdfjs` matching the `.gitignore` rules and therefore not being included when I `git add`-ed the `src/vendor/pdfjs` directory. I have updated this PR with the latest version of PDF.js, added the `pdfjs/build` directory and modified the `.gitignore` rules.

robertknight:
When testing this, I would suggest creating a build of the production extension and eyeballing a few PDFs that have recently been annotated with public annotations.

robertknight:
@sean-roberts - The missing vendor files in `src/vendor/pdfjs/build` have been added to Git. This is good for review again.

src/vendor/pdfjs/web/viewer.html
```diff
@@ -14,33 +14,36 @@
...
-http://sourceforge.net/adobe/cmap/wiki/License/
+Adobe CMap resources are covered by their own copyright but the same license:
+
+    Copyright 1990-2015 Adobe Systems Incorporated.
+
+See https://github.com/adobe-type-tools/cmap-resources
 -->
 <html dir="ltr" mozdisallowselectionprint moznomarginboxes>
   <head>
+<script async defer src="/public/embed.js"></script>
```
nickstenning: I reckon we probably only want/need `async` here, as that gives us some flexibility about when we instantiate what. (It's also what we recommended in [our own docs](https://h.readthedocs.io/en/latest/publishers/embedding/).)

tools/update-pdfjs
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+
```
nickstenning: It's always a good idea to make sure you catch unexpected errors and don't plough on through the script by adding (at least) `set -eu` at the top.

tools/update-pdfjs
```diff
@@ -0,0 +1,27 @@
...
+# Remove example content from viewer
+rm $DEST_DIR/web/*.pdf
+
+# Remove the check that the PDF being loaded is from the same origin as the
+# viewer
+sed -i '/^\s*validateFileURL(file)/d' $DEST_DIR/web/viewer.js
+
+# Add code to load the Hypothesis client when the viewer runs
+sed -i '/<head>/a <script async defer src="\/public\/embed.js"></script>' $DEST_DIR/web/viewer.html
```
nickstenning: Oops, I suppose my comment about the `defer` attribute should probably have gone here...

src/vendor/pdfjs/web/viewer.html
robertknight: Done. I have also changed the code that inserts the `<script>` to put it just before the closing `</head>` tag to match where it was placed previously and ensure that it always loads after the PDF.js scripts.

tools/update-pdfjs
robertknight: Done

tools/update-pdfjs
robertknight: Done. I have changed it here.

robertknight:
> On neither of these do I see rendered highlights, and on neither of these does clicking on annotations scroll to the appropriate part of the document.

Discussed in Slack, this was due to using a client build without https://github.com/hypothesis/client/pull/134

* Added `set -eu` to top of `tools/update-pdfjs`
* Changed the script in `tools/update-pdfjs` to insert the `<script>` tag using only async rather than defer and changed the sed command to insert it before the closing `</head>` tag where it was previously.
* Updated PDF.js to the current release (1.6.343) and tested with the two PDFs mentioned above

nickstenning:
Ok, looks good.

Issue #12, opened 2016-11-24
Anchoring failures after PDF.js update


robertknight:
After testing a bunch more PDFs against the recent PDF.js update, I found a case where annotations that were anchored in the previous version are now not anchored: https://github.com/hypothesis/browser-extension/pull/9

**Steps to reproduce:**

Using a production build of the Chrome extension and Hypothesis client >= v0.47.0:

1. Open https://www.jyu.fi/edu/laitokset/okl/koulutusala/vkluoko/tietopankki/tutkimusta/viittomakielinen_juhlajulkaisu_nettiversio.pdf
2. Activate Chrome extension
3. Watch the bucket bar counter in the bottom-right corner of the screen increment as annotations are anchored. Wait until it stops increasing.
4. Sort annotation list by location, scroll it to the bottom and click on it to scroll to the last annotation in the document

**Expected:** Annotation counter matches the number of annotations on the page as displayed in the sidebar (79 at the time of writing). Clicking on the annotation at the bottom of the sidebar scrolls the PDF viewer.
**Actual:** Annotation counter stops at a much smaller number (10 at the time of writing), clicking on the annotation at the bottom of the sidebar scrolls the PDF viewer 

**Notes**

Compare with Via which is running an older version of PDF.js. On Via, after the document opens, 79 annotations anchor - indicated by the counter in the bucket bar. With newer PDF.js, that counter only ever gets up to 10.

judell:
@robertknight this is a key finding that I want to include in the new testing harness, but I'm not seeing any annotations in the db for that url (or even a substring of it, like www.jyu.fi/edupsy).

judell:
@robertknight any chance you have that PDF lying around?

judell:
Update: Here is the report on what I've found so far, using an automatic method for comparing anchoring results (augmented with manual comparison): https://github.com/judell/h_puppeteer/blob/master/results2.md

tl;dr PDF.js v2 mostly works fine, as you'd expect since it is now what Firefox users are getting by default (by way of the bookmarklet) and we aren't hearing complaints. But the method used to produce that report, and the issues discussed in it, have yet to be reviewed.

PS: I have never tracked down the PDF that prompted the title of this now almost 3-year-old issue.

robertknight:
[...]

PR #13, opened 2016-11-24
Revert "Update PDF.js"

robertknight:
Reverts hypothesis/browser-extension#9 until the cause of the regression in https://github.com/hypothesis/browser-extension/issues/12 is found and fixed in the client code.

CC @nickstenning


nickstenning:
Okdoke.

PR #17, opened 2017-01-19
Adding Google Analytics staging/dev property to browser-extension

sean-roberts:
Part of the work for: hypothesis/product-backlog#117

Adds the client stage/dev property analytics tracking id


robertknight:
LGTM

PR #18, opened 2017-01-23
Identify extension appTypes

sean-roberts:
Instrumenting the appType identifier for chrome-extension

### DIFF
--- a/settings/chrome-dev.json
+++ b/settings/chrome-dev.json
@@ -8,5 +8,6 @@
 
   "googleAnalytics": "UA-26026798-6",
 
-  "browserIsChrome": true
+  "browserIsChrome": true,
+  "appType": "chrome-extension"
 }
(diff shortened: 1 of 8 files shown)


chdorner:
We should probably add `"appType": "firefox-extension"` to the `settings/firefox-*.json` settings.

sean-roberts:
Added

settings/chrome-dev.json
```diff
@@ -8,5 +8,6 @@
 
   "googleAnalytics": "UA-26026798-6",
 
-  "browserIsChrome": true
+  "browserIsChrome": true,
```
robertknight: I wondered where `browserIsChrome` is used at first, then I saw that it is used by the extension manifest template. In future we might want to separate out the settings that are only used when building the extension vs. those that are used by the client.

robertknight:
LGTM

PR #22, opened 2017-02-15
Add a script to automate creating a new extension release

robertknight:
- Add 'preversion' and 'postversion' scripts to support creating new
   client versions using the 'npm version' command.

 - Add a `tools/update-client` script which updates the client
   to the latest version on npm, runs the tests and creates a new
   tag release using the same version as the client package.

This automates steps 3-6 of the documentation at https://github.com/hypothesis/playbook/blob/master/docs/chrome-extension.md .

In future we might be able to avoid needing to update the extension for each client release by just having the extension load the boot script bundled with the client. Until then, this makes creating new extension releases a little bit easier.


package.json
```diff
@@ -41,6 +41,8 @@
     "whatwg-fetch": "^0.10.1"
   },
   "scripts": {
+    "postversion": "git push --follow-tags",
```
seanh: Cool, I didn't know about `--follow-tags`, but it's just what we want

tools/update-client
```diff
@@ -0,0 +1,25 @@
...
+# using the latest version of the client available on npm.
+
+set -eu
+
+LATEST_CLIENT_VERSION=$(npm view hypothesis version)
+
+# Update checkout
+git checkout master
+git fetch
+git reset --hard origin/master
```
seanh: I think this could delete uncommitted changes. Do you think it should run something like `git status --short` before it does any of these git commands? (And the script should crash with an error if there is any output from `git status --short`)

We could also make it do `git branch | grep '^\*'` and if the output isn't `* master` then abort, rather than changing the user's branch on them (potentially without them noticing)

robertknight: > I think this could delete uncommitted changes.

Yes, it will. Developers have to learn somehow :P

In all seriousness, I can change this to something a bit less brutal. `git status --name --untracked-files=no` should output an empty string if there are no uncommitted changes.

fatbusinessman: > `git status --name --untracked-files=no` should output an empty string if there are no uncommitted changes.

Can you also do `git diff --quiet HEAD` and check the exit code? I’ve given it a go and it *seems* to work whether the changes are in the index or not, and ignores untracked files in the same way. I’m aware I may well have missed some case where it fails, though.

robertknight: Thanks @fatbusinessman - I didn't know about the `--quiet` and `--exit-code` args to `git diff`.

tools/update-client
```diff
@@ -0,0 +1,30 @@
...
+LATEST_CLIENT_VERSION=$(npm view hypothesis version)
+
+# Check that working tree is clean
+if ! git diff --quiet HEAD ; then
+  echo "There are uncommitted changes in the working tree or index."
+  exit 1
+fi
+
+# Update checkout
+git checkout master
```
seanh: Hmm. I still think it'd be better to check whether they are on master and bail if they aren't, rather than quietly changing the user's branch on them which they may not expect or notice.

seanh:
Cool, the script is looking really nice and simple. I left one more comment about not changing branches, but I'll approve this and leave it up to you whether you want to make that change.

robertknight:
I hope you don't mind seanh, but I'm going to go ahead and merge this as-is and perhaps add a branch check later.

seanh:
Cool, no probs

PR #23, opened 2017-02-21
Update class name for config script tag in sidebar app

robertknight:
**This needs to be merged _before_ an extension containing https://github.com/hypothesis/client/pull/243 is released.**

https://github.com/hypothesis/client/pull/243 changes the client to use
the same class name, `js-hypothesis-config`, for config script tags in
both the sidebar app and host page.

This updates the extension's app.html template to use the expected name.


nickstenning:
:+1:

Issue #26, opened 2017-02-22
“Account settings” and “logout” menu items missing


fatbusinessman:
**Steps to reproduce**

* Visit any page using the browser extension while logged into Hypothesis
* Open up the user menu

**Expected behavior**

* The four options in the menu are “[username]”, “Account settings”, “Help” and “Log out”

**Observed behavior**

* The “Account settings” and “Log out” items are missing

![The user menu, missing two of its items](https://cloud.githubusercontent.com/assets/32775/23218248/ae7c4892-f913-11e6-910c-63373a1f9498.png)

**Other notes**

This happens when using the browser extension, but doesn’t happen when using the embedded client or [via](https://via.hypothes.is).

sheetaluk:
Fixed with https://chrome.google.com/webstore/detail/hypothesis-web-pdf-annota/bjfhmglciegochdpefhhlphglcehbmek

dwhly:
Verified.  Thank you Sheetal!

PR #27, opened 2017-02-22
Add authDomain to settings.

sheetaluk:
https://github.com/hypothesis/browser-extension/issues/26


robertknight:
In order to include these settings in `app.html`, `tools/template-context-app.js` will also need updating.

PR #29, opened 2017-02-24
Support setting the 4th component of the extension version

robertknight:
We use the first three components of the extension version for the
client version. It is useful to be able to set the 4th component [1] for
releases with extension-only changes.

We cannot set the 4th component in the package.json version field, but
we can set it in the tag name which is what is actually used to generate
the extension version.

This commit removes the auto-generated 4th component from the extension
version and instead extracts it from the tag version. With this change,
extension only releases can be made by leaving the version unchanged in
`package.json` but creating a signed tag such as `v0.59.0.1`.

[1] Chrome/Firefox/Edge extension versions may have up to 4 components.
    Any that are not set default to 0.

### DIFF
--- a/tools/settings.js
+++ b/tools/settings.js
@@ -20,8 +20,8 @@ process.stdout.on('error', err => {
  * rules:
  *
  * - If buildType is 'production' and the git state is not clean, throw an error
- * - Set the version number to X.Y.Z.W, where X.Y.Z is the last tagged release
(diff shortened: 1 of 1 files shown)


robertknight:
Nick reminded me that this is unnecessary because it should already work given the previous method of generating version numbers. We just need to avoid tagging new releases if there haven't been client changes.

PR #30, opened 2017-02-28
Unify the client's Sentry DSN

nickstenning:
It makes more sense to deduplicate all errors in Sentry using a single DSN. The client knows its own version number, which makes tracking down the source of errors easier.


robertknight:
LGTM

PR #31, opened 2017-03-01
Automate upload of Chrome extensions to Chrome Web Store

robertknight:
Use the [chrome-webstore-upload-cli](https://github.com/DrewML/chrome-webstore-upload-cli) tool to upload Chrome extensions to the Chrome Web Store automatically on pushes to master.

The stage extension is uploaded and published automatically. The prod extension is uploaded so the Chrome Web Store can start performing malware checks but not actually published. This gives us a chance to do any manual testing we want with the stage extension before hitting the Publish button in the Chrome Web Store's Developer Dashboard.

CHROME_WEBSTORE_{CLIENT_ID, CLIENT_SECRET, REFRESH_TOKEN} env vars have been added to the Travis CI config. Generating the refresh token is a PITA, so I have added a script to document and assist with the process if we ever need to do it again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,9 @@ addons:
     paths:
       # Upload all built extension packages
       - $(ls dist/*.zip dist/*.xpi | tr "\n" ":")
+deploy:
+  skip_cleanup: true
+  provider: script
+  script: ./tools/deploy
+  on:
+    branch: master
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "browserify": "^13.0.0",
     "chai": "^3.5.0",
     "check-dependencies": "^0.12.0",
+    "chrome-webstore-upload-cli": "^1.1.1",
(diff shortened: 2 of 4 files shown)


seanh:
Do you know if these refresh tokens expire?

Looks good to me. Can't test it without a new Chrome extension version to release. I guess the [docs for releasing the Chrome extension](https://github.com/hypothesis/playbook/blob/master/docs/chrome-extension.md) will need updating.

I think @sean-roberts should take a look at this one as well

robertknight:
> Do you know if these refresh tokens expire?

Google's documentation was unclear on that. In one of my own projects I did experience that a refresh token did stop working at some point long after it was issued (at least a year). Hence I don't think we need to automate the process of updating the refresh token but I have included a script as a form of documentation.

robertknight:
> I guess the docs for releasing the Chrome extension will need updating.

Yes, indeed.

PR #33, opened 2017-03-03
Add some npm link tips

seanh:
Add a couple of tips on using `npm link`. These come from <http://h.readthedocs.io/en/latest/developing/install/client/> which is going to be deleted

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,28 +71,12 @@ documentation for instructions on building the client in a development
 environment.
 
 **Tip**: If you get a **permission denied** error when running `npm link`
-you probably need to tell npm to install packages into a directory in your home
-directory that you have permission to write to. On linux:
-
-```sh
-npm config set prefix /home/<YOUR_USERNAME>/npm
-```
-
-On macOS:
-
-```sh
-npm config set prefix /Users/<YOUR_USERNAME>/npm
-```
-
-`npm` will now install executable files into `$HOME/npm/bin`, so add that
-directory to your `$PATH`.
+you probably need to fix npm's permissions. See
+[Fixing npm permissions](https://docs.npmjs.com/getting-started/fixing-npm-permissions).
 
 **Tip**: To **unlink** your dev browser extension from your dev client run
-`npm unlink hypothesis` in your browser extension directory:
-
-```sh
-npm unlink hypothesis
-```
+`npm unlink hypothesis` in your browser extension directory
+(see the [npm uninstall docs](https://docs.npmjs.com/cli/uninstall)).
 
 See [Building the extension](docs/building.md) for more information.
 
--- a/README.md
+++ b/README.md
@@ -70,6 +70,30 @@ built client from the client repository. Please consult the client's
 documentation for instructions on building the client in a development
 environment.
 
+**Tip**: If you get a **permission denied** error when running `npm link`
+you probably need to tell npm to install packages into a directory in your home
+directory that you have permission to write to. On linux:
+
+```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,6 +70,30 @@ built client from the client repository. Please consult the client's
 documentation for instructions on building the client in a development
 environment.
 
+**Tip**: If you get a **permission denied** error when running `npm link`
```
robertknight: I appreciate that this documentation was just moved from h.readthedocs.io but in general I'm wary of including generic documentation _for 3rd-party tools_. In this context for example, a search for 'npm link permission denied' turns up [this more detailed and up to date official documentation](https://docs.npmjs.com/getting-started/fixing-npm-permissions) on the subject, which we could just link to instead.

Thoughts?

seanh: This comes up now and then about our docs. Generally speaking I think there's a balance to be achieved between this idea of not duplicating third-party documentation, and making our own documentation user friendly. I agree that we don't want to document any third-party tool extensively when we could just link to its own docs. But I'm not averse to a couple of sentences explaining what a third-party tool is, what a command does, or the fix for a problem with a third-party tool that commonly occurs when following our instructions, etc. A lot of time I think we can make instructions unnecessarily difficult to follow by expecting people to find the third-party docs themselves, or even giving a potentially broken-in-future link to a page in the docs (and then requiring the reader to read that whole page and figure out what they need to do, even though they may not be familiar with or understand the third-party tool in question) rather than just giving the user the answer they need.

I certainly don't want our readers to have to google "npm link permission denied" to find the solution to an error message that can be reproduced by following our docs on a clean Ubuntu and that we could have supplied the solution to in our docs.

When we have a good, short third-party page to link to I think it's more grey. In this case the npm docs are more comprehensive so I'd be happy to replace the tip with a link to them.

(Although I have to say that their first suggestion of recursively changing the ownership of `/usr/local` from root sounds potentially dodgy to me, both in terms of possibly breaking multiuser systems as they point out, and also in terms of security.)

seanh: I've pushed a commit replacing the tips with links

robertknight: I think you make a lot of fair points.

I would say however that any explanations about the purpose and usage of common tools in a language's ecosystem should be brief - no more than a sentence or two, with a helpful link for people who haven't encountered them before. Even though we try to be careful about paring down the dependencies of the client and service, if we explained all the tools required to develop them in detail it would take up a lot of space and add a lot of noise for many readers.

If there is a particular feature of the tool which is less common that we recommend using (_npm link_ is a reasonable example), then a slightly more detailed explanation I think is OK.

> or even giving a potentially broken-in-future link to a page in the docs 

The other side is that any documentation we write, we also have to maintain. It could also get out of date when newer versions of the tool are released.

README.md
```diff
@@ -70,6 +70,14 @@ built client from the client repository. Please consult the client's
...
+you probably need to fix npm's permissions. See
+[Fixing npm permissions](https://docs.npmjs.com/getting-started/fixing-npm-permissions).
+
+**Tip**: To **unlink** your dev browser extension from your dev client run
+`npm unlink hypothesis` in your browser extension directory
+(see the [npm uninstall docs](https://docs.npmjs.com/cli/uninstall)).
```
robertknight: I wondered why this was `npm uninstall` and not `npm unlink` but it looks like `unlink` is currently an alias for uninstall and I couldn't find dedicated docs for it.

robertknight:
LGTM

PR #34, opened 2017-03-06
Use client boot script

robertknight:
**~~Depends on https://github.com/hypothesis/client/pull/270~~**

This PR removes the browser extension's own knowledge about the asset bundles used by the client (in its embed.js and app.html templates) in favor of using the client's boot script.

- Remove the embed.js template and instead use the boot script included with the client. In order to make the client load assets from the extension in this context, the `assetRoot` and `sidebarAppUrl` client config settings are used (see docs in https://github.com/hypothesis/client/pull/270). The config settings are passed to the client via a JSON script tag inserted into the page (this logic already exists and is not new in this PR).
- Change the app.html template to just include the client's boot script instead of listing out all the asset bundles used by the client. This mirrors a recent change to the service.
- Change the build process to put the client's assets at `/client/build` instead of `/public` in the generated extension. This means that `/client` mirrors the structure of the client package uploaded to S3 at `https://cdn.hypothes.is/hypothesis/X.Y.Z` which means that client boot script can assume the same package structure in both cases.

When testing this PR, you might see that the 'app.html' load is reported as failing in Chrome's dev tools, even though the sidebar app does actually load. This appears to be a bug relating to a recent change in Chrome to move iframes associated with extensions to separate processes. See comment in source code.


robertknight:
Marked as WIP because additional changes are still needed to get this working with the PDF viewer.

nickstenning:
As [observed](https://hypothes-is.slack.com/archives/public/p1489054763160534) [in Slack](https://hypothes-is.slack.com/archives/public/p1489064612079665), we would like to get to the point where this repository doesn't make any assumptions about the internal structure of the client repository (like the existence of a `build/` directory).

Unfortunately there are a couple of things that stand in the way of this:

1. The client still generates a `manifest.json` file (which isn't needed) which has to be manually removed from the package otherwise the Chrome Web Store complains.
2. In development `node_modules/hypothesis` is a symlink to the repo checkout, and we want to avoid copying *everything* in the repo into the extension. This is straighforwardly solved by instead doing `cd node_modules/hypothesis && npm pack`, but unfortunately *that* kicks off a client build, and by the time that's done everybody has gone back to their parents place for the Christmas holiday.

So, in order to clean this up, we need to:

- kill off the `manifest.json` file in the client
- make the client build fast, OR
- wait for npm v5 to be release, which apparently no longer runs `prepublish` tasks when running `npm pack`

For now, I'm merging this.

PR #39, opened 2017-06-16
Don't do git checkout or pull from a script

seanh:
The docs for releasing a Chrome extension say to checkout and pull the
latest master branch manually.

The script as its currently written assumes that the user's master
branch is setup as a tracking branch of the hypothesis organization's
master branch on GitHub, which while a common setup isn't always the
case (it's not the case for me, for example, so the script crashes for me).

For example it's common to fork a github repo and then clone your fork.

### DIFF
--- a/tools/update-client
+++ b/tools/update-client
@@ -13,9 +13,17 @@ if ! git diff --quiet HEAD ; then
   exit 1
 fi
 
-# Update checkout
-git checkout master
-git pull
+# Fetch commits from main repo and write commit ref to `.git/FETCH_HEAD`
+git fetch --quiet https://github.com/hypothesis/browser-extension.git master
+
+# Check that HEAD is the latest commit on the
+# github.com/hypothesis/browser-extension repo.
+remote_head=$(cat .git/FETCH_HEAD | cut -f1)
+local_head=$(git rev-parse HEAD)
+if [ $remote_head != $local_head ]; then
+  echo "Current commit does not match head of hypothesis/browser-extension.git:master"
+  exit 1
+fi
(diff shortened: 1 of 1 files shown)


seanh:
If we want to make it more automated (so that the user doesn't have to checkout and pull master manually) I wonder if the script should make git clone of `https://github.com/hypothesis/browser-extension` in a temporary directory, use that, and then delete it?

robertknight:
As discussed on Slack, the original motivation was to prevent developers forgetting to checkout the right branch and update it to match the remote before they create a new version.

The best solution will be to have Jenkins release the extension when a new client release is made.

An alternative that could work in the meantime is just to verify that the current HEAD matches the current HEAD of the `hypothesis/browser-extension` remote, which can be done using something like:

```sh
# Fetch commits from main repo and write commit ref to `.git/FETCH_HEAD`
git fetch --quiet https://github.com/hypothesis/browser-extension.git master
# Compare heads 
remote_head=$(cat .git/FETCH_HEAD | cut -f1)
local_head=$(git rev-parse HEAD)

if [ $remote_head != $local_head ]; then
  echo "Current commit does not match head of `hypothesis/browser-extension.git:master`"
  exit 1
fi
```

seanh:
Ok @robertknight , I've replaced it with a version containing your script. I tested this by commenting out the lines before that check for a dirty working tree, and commenting out the lines after that actually run the npm and other commands, and then testing that on this branch it exits with `Current commit does not match head of hypothesis/browser-extension.git:master` but on master it does not. So I think this'll work.

PR #40, opened 2017-06-16
Fix postversion git push command

seanh:
Fix the `git push` command that `npm version` runs after making a new
version.

The command as it was assumed that the user's git repo's `origin` remote
is https://github.com/hypothesis/browser-extension, which is not always
the case. For example it's common to fork an open source repo to another
github (or bitbucket etc) account and then clone your own fork locally.


robertknight:
LGTM.

PR #41, opened 2017-06-23
Fix client failing to load in Firefox WebExtension

robertknight:
This is a set of three fixes for the Firefox WebExtension which, together with https://github.com/hypothesis/client/pull/460 , enables the extension to work in the current stable version of Firefox (v51).

See individual commit messages for details.

**Steps to test:**

1. Checkout https://github.com/hypothesis/client/pull/460 in the client repo and build the client.
2. Checkout this PR in the browser-extension repo
3. Build the Firefox dev extension `make SETTINGS_FILE=settings/firefox-dev.json`, making sure that you are using the client build from (1)
4. In the stable version of Firefox, go to "about:debugging#addons", click "Load Temporary Addon" and select the "manifest.json" file from the "build" directory.

### DIFF
--- a/src/common/help-page.js
+++ b/src/common/help-page.js
@@ -63,8 +63,9 @@ function HelpPage(chromeTabs, extensionURL, browserName_) {
       url:  extensionURL('/help/index.html' + params + '#' + helpSection),
     };
 
+    // Add `openerTabId` property to associate the help page tab with the
+    // current tab. This property is not supported in Firefox.
     if (browserName_() !== 'firefox') {
-      // We are in Firefox, which does not support the `openerTabId` parameter.
       tabOpts.openerTabId = tab.id;
     }
 
--- a/src/common/help-page.js
+++ b/src/common/help-page.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var browserName = require('./browser-name');
 var errors = require('./errors');
 
 /* A controller for displaying help pages. These are bound to extension
@@ -10,7 +11,9 @@ var errors = require('./errors');
  *   to the file inside the chrome extension. See:
  *   https://developer.chrome.com/extensions/extension#method-getURL
  */
-function HelpPage(chromeTabs, extensionURL) {
+function HelpPage(chromeTabs, extensionURL, browserName_) {
+  browserName_ = browserName_ || browserName;
+
   /* Accepts an instance of errors.ExtensionError and displays an appropriate
    * help page if one exists.
    *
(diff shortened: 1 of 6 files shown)


seanh:
On this branch I'm seeing the `/welcome/` tab open when I install the Firefox extension, but the sidebar doesn't automatically open on the page. Also happens on master though so I'll assume it's a separate issue.

seanh:
@robertknight How do I reproduce the help page issue that the first commit fixes?

robertknight:
> @robertknight How do I reproduce the help page issue that the first commit fixes?

On master (client & browser extension), click to activate the extension, it will fail and show a "!" mark. Then click the "!" again and instead of opening the help page it will do nothing. In this branch, clicking the "!" will open the help page. You can force failure by modifying `injectIntoTab` to return a Promise rejected with some `Error`.

seanh:
Confirmed the first commit: on master open a local HTML file (`file:///...*.html`) in Firefox, and try to activate the Firefox extension. It displays a `!`. Click again. Nothing happens. It should open the extension's built-in `/help` page. On this branch it does open the help page. (This works just with this branch of the extension and the master branch of the client.)

seanh:
Confirmed the bug that the second commit should fix: after downloading Firefox 52 from <https://ftp.mozilla.org/pub/firefox/releases/>, on master browse to a page that has no annotations and then to a page that does have annotations - the annotation count in the extension's browser button does not update.

I can't reproduce this bug in Firefox 54 - the count does update (even with the master version of the client and the extension).

However, for me **the bug still seems present even on this branch** with Firefox 52.

seanh:
Confirmed the third commit: on master, trying to launch the client from the Firefox extension fails with `Injecting Hypothesis sidebar Error: Script returned non-structured-clonable data` in the extension's background window. On this branch of the browser extension that error is gone but I get `chrome.contentSettings is undefined` in the background window instead and the extension still fails to launch. But if I then checkout the <https://github.com/hypothesis/client/pull/460> branch of the client as well and then re-build the extension, then launching the sidebar does work.

src/common/help-page.js
```diff
@@ -55,11 +58,17 @@ function HelpPage(chromeTabs, extensionURL) {
...
-    chromeTabs.create({
+    var tabOpts = {
       index: tab.index + 1,
       url:  extensionURL('/help/index.html' + params + '#' + helpSection),
-      openerTabId: tab.id,
-    });
+    };
+
+    if (browserName_() !== 'firefox') {
+      // We are in Firefox, which does not support the `openerTabId` parameter.
```
seanh: Should this comment read "We are _not_ in Firefox"?

robertknight: Gah, indeed. Fixed!

PR #42, opened 2017-06-26
Enable ES6 usage

robertknight:
Enable use of ES6 language features using Babelify.

Since the code in this repository only needs to target reasonably recent
versions of Chrome and Firefox, use [babel-preset-env](https://github.com/babel/babel-preset-env) to control which
ES6 => ES5 transformations are applied based on the browser versions we
want to support.

The versions are initially set conservatively to the minimum Firefox
version that has sufficiently complete WebExtensions support (48) and
Chrome circa ~2 years ago. We can bump these in future when we get data
that gives us a better idea of what versions our users are actually
using.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,10 @@
+{
+  "presets": [
+    ["env", {
+      "targets": {
+        "firefox": "48",
+        "chrome": "43"
+      }
+    }]
+  ]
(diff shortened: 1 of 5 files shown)


seanh:
:+1: Thanks for taking care of this

PR #45, opened 2017-07-06
Add support for "#annotations:query:<query>" fragments

robertknight:
Support direct link queries which specify a search query for annotations instead of an ID.
    
This enables bouncer links with a query, eg. https://hyp.is/go?url=https://example.com&q=user:robertknight to work when bouncer detects that the user has a suitable browser extension installed.


seanh:
Looks great!

Issue #46, opened 2017-07-16
Why does Chrome browser extension ask for geolocation, camera & microphone settings access?


billfitzgerald:
While adding the Chrome plugin this morning from the Chrome store, I received the message below that states that the extension needs to be able to change website's access to geolocation settings, camera, and microphone.

![why](https://user-images.githubusercontent.com/1602849/28249064-0e70e814-6a04-11e7-9615-902c5694169c.png)

I'm assuming that this message appears because of how these permissions are bundled, but from a functional place,  

I'm assuming that this is a non-issue related more to how the Chrome store auto-generates messages based on what permissions are requested, but given that the message pops up on install, I'd love to know more about:

- why does the annotation plugin need to be able to access geolocation, camera, or microphone settings?
- if the plugin does need to be able to adjust access to these settings, how are users notified that their defaults are getting changed?
- does any data collected via mic/camera/geolocation get sent back to hypothes.is servers (in addition to the originating web site)?

Thanks for any info you can provide here.

Also, some general technical details: I'm running Ubuntu Linux 16.04, latest Chrome - glad to share additional technical details privately if they are helpful/relevant.

robertknight:
> I'm assuming that this is a non-issue related more to how the Chrome store auto-generates messages based on what permissions are requested

Hello @billfitzgerald - you are completely correct. This is related [to a change](https://github.com/hypothesis/browser-extension/blob/7e0823d42db2140d6d99b7d0e8009189031155ff/src/common/install.js#L59) that the extension makes to your cookie whitelist (adding "https://hypothes.is" to an always-allow whitelist, even if third party cookies are otherwise blocked) in order to enable logging in to the service from the sidebar to work.

You can see the change at the bottom of this screenshot:

<img width="711" alt="screenshot 2017-07-17 09 21 29" src="https://user-images.githubusercontent.com/2458/28260109-739a4b7c-6ad1-11e7-9559-e9acc09e4f73.png">

The extension **does not read or modify your camera, geolocation or microphone settings or use these devices in any way**. Unfortunately these permissions are part of the same set of permissions that allows the software to change the cookie whitelist.

We are currently in the process of changing the way authentication works in the client and should be able to remove the need to ask for this permission after that is done.

robertknight:
I'm going to leave this open so we can close this once the client & service's OAuth implementation is complete.

billfitzgerald:
@robertknight - thank you! That's great to know, both re current status and future plans.

[...]

PR #51, opened 2017-09-06
Stop adding h domain to cookie whitelist during install

robertknight:
As of v1.39, the Hypothesis client detects when third party cookies are
blocked and will use OAuth rather than cookie authentication in that
case. Therefore, we can stop asking for the "contentSettings" permission which produces a scary warning (see hypothesis/browser-extension#46) and remove the code that modifies these browser settings.

There is a potential issue in that the detection of cookie blocking checks reading/writing cookies against the browser extension's domain (`chrome-extension://{id}`) whereas the actual cookies are set for the hypothes.is domain. However, the "block third party cookies" setting in Safari, Firefox and Chrome treats both as third party cookies and so the "can I set cookies" check that the client works as expected.

Fixes hypothesis/browser-extension#46
Fixes hypothesis/product-backlog#359

----

Testing:

1. Remove *all* Hypothesis extensions from Chrome
1. Build this branch using `make SETTINGS_FILE=settings/chrome-prod.json`  and add it to Chrome as an unpacked extension
1. Go to chrome://settings/content/cookies, enable third party cookie blocking and verify that no Hypothesis domains appear in the "Allow" list.
1. Go to https://example.org , activate the extension and click "Log in". This should trigger a pop-up window instead of showing the inline login form. The pop-up may flash and disappear immediately.

Note that after signing in in the pop-up window, nothing happens in the client because a locally-built client currently has a different ID than the one installed from the Chrome web store, so the browser will block the cross-window message that contains the auth code from reaching the client. I'm going to fix this separately.

### DIFF
--- a/src/chrome/manifest.json.mustache
+++ b/src/chrome/manifest.json.mustache
@@ -44,7 +44,6 @@
   "permissions": [
     "<all_urls>",
 
-    "contentSettings",
     "storage",
     "tabs"
   ],
--- a/src/common/install.js
+++ b/src/common/install.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var HypothesisChromeExtension = require('./hypothesis-chrome-extension');
-var settings = require('./settings');
 
 var browserExtension = new HypothesisChromeExtension({
   chromeExtension: chrome.extension,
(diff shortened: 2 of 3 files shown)


tilgovi:
*Yes*!

PR #57, opened 2017-11-06
Use fixed IDs for locally built stage and prod extensions

robertknight:
_Context: I'm currently updating the client's [development docs](http://h-client.readthedocs.io/en/latest/developers/developing/#setting-up-a-client-development-environment) and addressing the fact that the [running from a browser extension](http://h-client.readthedocs.io/en/latest/developers/developing/#running-the-client-from-the-browser-extension) steps don't "just work" after the OAuth changes_.

----

By default, locally built Chrome extensions get system-dependent IDs which differ from the IDs that are assigned when the extension is published in the Chrome Web Store. The IDs are part of the URLs of the extension's assets (`chrome-extension://{ID}/{path}`), including the url of the sidebar app (`/client/app.html`) which makes the request to the h service's authorization endpoint.

This means that a separate OAuth client would need to be registered in the h service for each user who wanted to build the prod/stage extensions locally. This is a) inconvenient and b) not possible for non-Hypothesis staff.

This PR populates the [key field](https://developer.chrome.com/apps/manifest/key) of the extension manifest at build time so that locally built prod & stage extensions have the same IDs ("bjfhmglciegochdpefhhlphglcehbmek" and "lhieaifenniokmcmfdcgabhbnjmelchg" respectively) as those from the Chrome Web Store, allowing the existing OAuth clients registered in the production "h" service to work.

A downside of this change is that it is not possible to have a locally built prod or stage extension installed at the same time as the "real" versions from the Chrome Web Store, because the IDs would conflict. We could fix this by enabling users to [register OAuth clients themselves](https://github.com/hypothesis/product-backlog/issues/376) in future.

----

**Testing:**

1. Uninstall prod/stage extensions from Chrome Web Store
2. Build prod extension from this PR
3. Check that the ID shown in chrome://extensions is "bjfhmglciegochdpefhhlphglcehbmek"
4. Build stage/QA extension from this PR
5. Check that the ID shown in chrome://extensions is "lhieaifenniokmcmfdcgabhbnjmelchg"

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -41,9 +41,13 @@ Once you've built the extension, you will be able to load the `build/` directory
 as an unpacked extension:
 
 1.  Go to `chrome://extensions/` in Chrome.
-2.  Tick **Developer mode**.
-3.  Click **Load unpacked extension**.
-4.  Browse to the `build/` directory where the extension was built and select it.
+2. If you used the `chrome-prod.json` settings file to build a production
+   extension, you will need to **remove** the "real" production extension from
+   Chrome before loading your locally built one or create a new Chrome profile
+   without the real one installed.
+3.  Tick **Developer mode**.
+4.  Click **Load unpacked extension**.
+5.  Browse to the `build/` directory where the extension was built and select it.
 
 Your extension should be working now! Remember that if you built a development
 extension it will point to a Hypothesis service running on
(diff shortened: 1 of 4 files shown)


seanh:
> A downside of this change is that it is not possible to have a locally built prod or stage extension installed at the same time as the "real" versions from the Chrome Web Store

That seems annoying. Could we fix the IDs of locally built stage and prod extensions to something _other_ than the real stage and prod extensions, so that they wouldn't conflict? And then add these new fixed IDs to the production h as auth clients.

robertknight:
Yes, I _think_ that should be possible. We would need to:

1. Create a couple of unlisted extensions in the Chrome Web Store for this purpose, install them once and get the "key" property from the installed manifest.
2. Provide some way to set the "key" value in the manifest when doing a local prod or staging build but not when doing a real prod / stage build, since the IDs will be provided by the Chrome Web Store in that case. We could simply provide a separate settings file (`chrome-dev-with-prod-h.json`?)

There probably isn't that much value doing this for the stage extension initially, so I would create just one dummy entry in the CWS for a "Development prod extension". Thoughts?

seanh:
Slack discussion: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1510075254000405

As discussed gonna merge this as-is now, it's a pragmatic solution and we can make it more perfect later if we want to

PR #65, opened 2018-03-10
Disable badge by default

hmstepanek:
Disable the badge by default and on firefox disable it permanently
(as there is no setting to enable it). In order to do this we need
to add a new badgeEnable state param to the store, otherwise any
previously installed extension settings will override the default
settings of the latest version and the badge won't be disabled.
The badge is being changed to an opt in setting because it accounts
for 90% of our transactions and is a scallabilty problem. Every
page the user browses to when the badge is enabled causes a
transaction to es and our db.

### DIFF
--- a/src/chrome/lib/options.js
+++ b/src/chrome/lib/options.js
@@ -3,14 +3,16 @@
 function saveOptions() {
   chrome.storage.sync.set({
     badge: document.getElementById('badge').checked,
+    badgeEnable: document.getElementById('badge').checked,
   });
 }
 
 function loadOptions() {
   chrome.storage.sync.get({
     badge: true,
+    badgeEnable: false,
   }, function(items) {
-    document.getElementById('badge').checked = items.badge;
+    document.getElementById('badge').checked = items.badgeEnable;
   });
 }
 
--- a/src/common/hypothesis-chrome-extension.js
+++ b/src/common/hypothesis-chrome-extension.js
@@ -285,14 +285,17 @@ function HypothesisChromeExtension(dependencies) {
   function updateAnnotationCountIfEnabled(tabId, url) {
     if (!chromeStorage.sync) {
       // Firefox < 53 does not support `chrome.storage.sync`.
-      state.updateAnnotationCount(tabId, url);
+      // If there are no setting options for dissabling the query
+      // for the number of annotations on a page just don't run it.
+      // state.updateAnnotationCount(tabId, url);
       return;
     }
 
     chromeStorage.sync.get({
       badge: true,
+      badgeEnable: false,
     }, function (items) {
-      if (items.badge) {
+      if (items.badgeEnable) {
         state.updateAnnotationCount(tabId, url);
       }
     });
--- a/tests/common/hypothesis-chrome-extension-test.js
+++ b/tests/common/hypothesis-chrome-extension-test.js
@@ -61,7 +61,7 @@ describe('HypothesisChromeExtension', function () {
   beforeEach(function () {
     fakeChromeStorage = {
       sync: {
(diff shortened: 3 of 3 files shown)


segdeha:
This is not reflective of any product decision we’ve made nor required to mitigate scaling issues. Please do not merge this until we’ve had a chance to discuss.

hmstepanek:
I know we haven’t made any sort of official decision yet. Figured I’d have this ready if this is the route we decide to take even temporarily until we have another solution. Don’t want to cause any upset by opening this, sorry. 😕

segdeha:
Thanks for clarifying! We'll talk about it on Monday. Thanks, @hmstepanek!

seanh:
If disabling the badge temporarily we might want to disable it server-side instead, by hardcoding the API endpoint to just return 0 (at the highest possible level: Cloudflare? Nginx? Prevent it from even hitting our app servers at all if possible). Reason is that we can just turn it on again right away server-side, without having to do a browser extension release and wait for it to get released out to users. Looks like this was [previously done in nginx](https://github.com/hypothesis/playbook/commit/51b41705be7025da4ed4673f3a3b7d16fac351b0) ([and later reverted](https://github.com/hypothesis/playbook/commit/5e3699cec953c55bfc6065705a3c405c5634b513)). That commit would no longer be valid today though - looks like the files in the playbook repo have moved around too much.

It would be good to know how to quickly temporarily disable the badge server-side, so that we can do so in an emergency if we need to.

Of course disabling it in the browser extension itself would be the best long-term as it means the requests don't even get sent

hmstepanek:
Yes we can just return 0 in the backend but I think it's best as you said there at the end to not hit our server at all. Plus I think this at least satisfies the need for some customers to still keep it if they wish (like Dan for example). It's not disabling it permanently (except firefox--it looks like we don't have an option configuration for it on firefox or something..?) it's simply turning it from an opt-out to an opt-in.

judell:
It would be good to have a discussion about better ways to solve the problem the badge addresses. The problem starts with the fact that the extension is by default inactive, and must be activated on a per-tab basis. If the extension were always benign, it might be OK to keep it active. But it isn't always benign. Important sites that it currently interferes with include GMail and Slack. 

So, if it's up to the user to activate the extention on a per-tab basis, there are two cases:

1. There are no annotations, and I want to annotate

2. There /are/ annotations, I want to see them.

In case 2, without the badge, a user who wants to be alerted to the existence of annotations has to remember to check every tab. That's unrealistic.

So we have the badge, which signals case 2, but is horribly expensive. Is there a less-expensive way to send that signal? And/or is there a non-disruptive way to keep the extension always on?

dwhly:
Agree w/ folks here and in other channels that this is something that warrants a thoughtful product discussion.  @ajpeddakotla can you help bring us together on this?

Separately, to @seanh's point above:

> it would be good to know how to quickly temporarily disable the badge server-side, so that we can do so in an emergency if we need to.

Yes!

ajpeddakotla:
Peeps, I'm going to wait until this sprint is done, so that we can focus on our current sprint priorities. I'll schedule a convo for next week that's open to the team, so we can figure out how best to proceed. Let's hold off on further discussion until then please.

robertknight:
As input to the engineering aspect of the discussion next week, I did some prototyping and design explorations around a standalone service to support the badge. The prototype's code and design document can be found at https://github.com/hypothesis/badger.

seanh:
> Peeps, I'm going schedule a convo for next week that's open to the team. 
Let's  wait until next week to discuss this, so that we can focus on our current sprint priorities. 

> Please hold off on further discussion until then. 

+1, thanks Arti

hmstepanek:
I think if you were to temporarily disable the badge you would block the request with nginx.

PR #73, opened 2018-06-08
Guard more manifest keys that as Chrome-only

tilgovi:
These keys appear to not work in Firefox.

### DIFF
--- a/src/chrome/manifest.json.mustache
+++ b/src/chrome/manifest.json.mustache
@@ -2,7 +2,9 @@
   "name": "Hypothesis - Web & PDF Annotation",
   "short_name": "Hypothesis",
   "version": "{{ version }}",
+  {{#browserIsChrome}}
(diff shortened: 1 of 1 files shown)


robertknight:
Thanks @tilgovi! Upstream bugs:

`version_name` field - https://bugzilla.mozilla.org/show_bug.cgi?id=1380219
`offline_enabled` - Couldn't find a mention outside of the [Chrome dev docs](https://developer.chrome.com/extensions/manifest/offline_enabled)
`externally_connectable` - https://bugzilla.mozilla.org/show_bug.cgi?id=1319168

The Firefox extension looked somewhat borked when I tried it in Nightly just now, but that's unrelated to the changes in this PR.

robertknight:
The lack of `externally_connectible` support could be a problem for getting https://hyp.is links working.

PR #77, opened 2018-08-13
Upgrade vulnerable deps with `npm audit fix --force`.

jmcarp:
Same deal as https://github.com/hypothesis/h/pull/5222. There are multiple packages upgraded with major version changes, but tests are passing. It would probably be good to run these checks via ci if that's not happening already.


robertknight:
Thanks for this @jmcarp. We do in fact have GitHub's vulnerability checks enabled for this repository.

PR #80, opened 2018-10-09
Revert "Make return value of boot script serializable"

robertknight:
Appending this token to the end of `boot.js` causes a syntax error with the
latest version of the client since the structure of the code changed.

Additionally the hack does not appear to be required any more (tested with Firefox 64).

This reverts commit 2674d22cbfda70627ba225f99af815a457f1ab19.


robertknight:
This was caught by Mozilla's bundle validation rather than our own tests (!) - https://addons.mozilla.org/en-US/developers/upload/5ce50f5387114b4b9f0addddb7afdd77#general-tests

Additionally it seems that Chrome's own validation doesn't catch this issue.

PR #120, opened 2019-02-26
Run tests under headless Chrome instead of PhantomJS

robertknight:
Since this project is a browser extension for modern browsers, in particular Chrome and Firefox, it makes sense to run the tests under a headless modern browser too.

I intend to make this change for the client as well, but the browser extension is an easier place to start since there is only the Travis CI build to deal with, not Jenkins as well.

A side benefit is that we can remove some polyfills that were only necessary under PhantomJS.


judell:
Excellent news, thanks for doing this.

PR #141, opened 2019-04-12
Support direct links to groups

robertknight:
Support the `#annotations:group:<ID>` syntax used to direct-link to
groups.

As part of this, refactor the code so that the knowledge about the
different kinds of direct-link query that can be used only exists in
one place, `direct-link-query.js`, which is responsible for mapping the
query to additional client configuration.

Also add some comments about why the browser extension has to deal with
direct-links itself in the first place.

----

To make it harder to overlook this in future if we change the direct linking syntax again, we could rework the overall flow as follows:

1. User navigates to $BOUNCER_URL/$QUERY
2. Bouncer detects the extension and sends it a message indicating the URL to open in the current tab and additional client configuration to pass to the client
3. The browser extension responds to this message by loading the provided URL and passing the given config to the client

This would avoid the browser extension needing to know the details of the direct-linking syntax.

### DIFF
--- a/src/common/direct-link-query.js
+++ b/src/common/direct-link-query.js
@@ -38,7 +38,9 @@ function directLinkQuery(url) {
     return { query };
   }
 
-  var groupMatch = url.match(/#annotations:group:(.*)$/);
+  // Group IDs (and other "pubids" in h) are a subset of ASCII letters and
+  // digits.
+  var groupMatch = url.match(/#annotations:group:([A-Za-z0-9]+)$/);
   if (groupMatch) {
     return { group: groupMatch[1] };
   }
--- a/src/common/direct-link-query.js
+++ b/src/common/direct-link-query.js
@@ -1,30 +1,46 @@
 'use strict';
 
 /**
+ * Subset of the client configuration which causes the client to show a
+ * particular set of annotations automatically after it loads.
+ *
+ * See https://h-client.readthedocs.io/en/latest/publishers/config/#config-settings
+ *
  * @typedef {Object} Query
- * @property [string] id
- * @property [string] query
+ * @property {string} [annotations] - ID of the direct-linked annotation
+ * @property {string} [query] - Filter query from the sidebar
+ * @property {string} [group] - ID of the direct-linked group
  */
 
 /**
- * Extracts a direct-linked annotation ID or query from the fragment of a URL.
+ * Extracts the direct-linking query from the URL if any.
  *
- * @param {string} url - The URL which may contain a '#annotations:<ID>'
- *        or '#annotations:query:<query>' fragment.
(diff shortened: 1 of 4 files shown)


src/common/direct-link-query.js
```diff
@@ -1,24 +1,35 @@
...
 /**
+ * Subset of the client configuration which causes the client to show a
+ * particular set of annotations automatically after it loads.
+ *
+ * See https://h-client.readthedocs.io/en/latest/publishers/config/#config-settings
+ *
  * @typedef {Object} Query
- * @property [string] id
- * @property [string] query
+ * @property {string} [annotations] - ID of the direct-linked annotation
```
robertknight: The property name changed from `id` to `annotations` here so that the resulting object can be merged directly into the client's configuration. This avoids the rest of the extension needing to know about the details of direct linking, beyond whether there is a direct link or not.

tests/common/direct-link-query-test.js
```diff
@@ -11,14 +11,21 @@ describe('common.direct-link-query', () => {
   it('returns the ID if the URL contains a #annotations:<ID> fragment', () => {
```
hmstepanek: Might be good to clarify ID is the annotation id or even the direct-linked annotation id.

tests/common/direct-link-query-test.js
```diff
@@ -11,14 +11,21 @@ describe('common.direct-link-query', () => {
...
 
-  it('returns the query if the URL contains a #annotations:<query> fragment', () => {
+  it('returns the query if the URL contains a #annotations:query:<query> fragment', () => {
     var url = 'https://example.com/#annotations:query:user%3Ajsmith';
     assert.deepEqual(directLinkQuery(url), {
       query: 'user:jsmith',
     });
   });
+
+  it('returns the group if the URL contains a #annotations:group:<id> fragment', () => {
```
hmstepanek: For consistency can we upper case the ID in `#anotations:group:<id>`? Perhaps "returns the **direct-linked group id** if the URL..."

src/common/direct-link-query.js
```diff
@@ -27,6 +38,11 @@ function directLinkQuery(url) {
     return { query };
   }
 
+  var groupMatch = url.match(/#annotations:group:(.*)$/);
```
hmstepanek: This should be matching url-safe-base64 identifiers rather than `.*`.

robertknight: I had another look at pubids in h. It turns out they are different to annotation IDs. They are currently a subset of alphanumeric characters designed to avoid certain (English-language) readability issues: https://github.com/hypothesis/h/blob/b4944f117232fd4eb497b72e1ddd4f5879e974fe/h/pubid.py#L6

PR #142, opened 2019-04-12
Convert proxyquire to mockable imports

robertknight:
This converts mocking in tests from proxyquire over to the mockable-imports Babel plugin. See https://github.com/hypothesis/client/pull/1061 for the rationale and explanation.

In the context of the browser extension, this has the following benefits:

- Consistency with the Hypothesis client going forwards
- Avoids several of the footguns that Proxyquire has (see client PR and linked posts)
- Will make some modernization/simplification of the code easier in future. For example if we decide to convert require to ES6 imports

This was mostly straightforward except for `src/common/install.js`, where some code had to be moved from the top-level of the module into a function which, in tests, is called explicitly. This is because, with this new approach to mocking, mocks are not active when the module is initially evaluated (this is also the case with `patch` in Python for example, which this new approach is more similar to).


src/common/browser-action.js
```diff
@@ -46,6 +44,8 @@ function _(str) {
  * the badge state.
  */
 function BrowserAction(chromeBrowserAction) {
+  var buildType = settings.buildType;
+
```
robertknight: This was moved here so that the current value of `settings.buildType` at the time `new BrowserAction` is called is used.

tests/common/hypothesis-chrome-extension-test.js
```diff
@@ -111,6 +109,7 @@ describe('HypothesisChromeExtension', function() {
       removeFromTab: sandbox.stub().returns(Promise.resolve()),
     };
     fakeErrors = {
+      AlreadyInjectedError: function AlreadyInjectedError() {},
       shouldIgnoreInjectionError: function() {
```
robertknight: This mock was missing. It happened to work because proxyquire "calls through" to the original module for exports that are missing in the mock.

tests/common/hypothesis-chrome-extension-test.js
```diff
@@ -121,27 +120,26 @@ describe('HypothesisChromeExtension', function() {
...
-        './help-page': createConstructor(fakeHelpPage),
-        './browser-action': createConstructor(fakeBrowserAction),
-        './sidebar-injector': createConstructor(fakeSidebarInjector),
-        './errors': fakeErrors,
-        './settings': {
-          serviceUrl: 'https://hypothes.is/',
-        },
-      }
-    );
+    FakeTabState.states = TabState.states;
```
robertknight: Same deal as above, the `states` property was missing from the mock but proxyquire used to call through to the original.

src/common/install.js
```diff
@@ -2,20 +2,28 @@
...
-    return chrome.extension.getURL(path);
-  },
-  isAllowedFileSchemeAccess: function(fn) {
-    return chrome.extension.isAllowedFileSchemeAccess(fn);
-  },
-});
+var browserExtension;
 
-browserExtension.listen(window);
+function init() {
```
robertknight: The code here has been moved into an init function which is also exported to deal with the main drawback of the new approach, which is that you can't mock code that runs during the initial evaluation of a module. See the "Drawbacks" section of https://github.com/hypothesis/client/pull/1061.

I actually think this might be a good thing by encouraging imports which are side-effect free. However, one place where it will be a problem are modules which are entry-points for the app, like this one. There is a relatively simple workaround though which is to wrap the code in an init function which is run immediately in the real app but not during tests.

PR #169, opened 2019-06-19
Update main h logo with new icon

### LINKED ISSUE #168: Update the extension logo that appears in the extension tab
While the app's mini logo has been updated, we also need to update the main logo for the extension that appears in the chrome://extensions/ tab.

h logo svg
https://gist.github.com/jaredpdesigns/626aa057417f7c82bfe352557cc9c93a

*note: we made need to convert to a png*

Related to: hypothesis/product-backlog#1006

LMS007:
Update the main icon in the chrome:extensions tab / store to the new artwork. Fixes #168


hmstepanek:
Before:
image

After:
image

Before:
image

After:
image

hmstepanek:
@LMS007 can you provide some context here as to what we're trying to achieve specifically? It looks like we are just updating the h logo but I'd like to understand where these are used currently and have before and after captures of each. (I tried to do that a moment ago but I felt like I was doing a pretty bad job of it.) I'm also wondering, these logos appear identical except for the size-is this something where we could have one file and specify size it in css somewhere?

hmstepanek:
Before:
image

After:
image

LMS007:
> @LMS007 can you provide some context here as to what we're trying to achieve specifically? It looks like we are just updating the h logo but I'd like to understand where these are used currently and have before and after captures of each. (I tried to do that a moment ago but I felt like I was doing a pretty bad job of it.) I'm also wondering, these logos appear identical except for the size-is this something where we could have one file and specify size in css somewhere? I'm also wondering why these are png's and not svg's (I realize they were that way before this PR just trying to understand as I haven't done much work in the browser extension and I'm not familiar with the current state of it.)?

I'll see if SVG (with scaling) is do-able -- it should be. Not 100% sure where the sizes are used either. Look into that.

dwhly:
We should check the icon out in the new MacOS dark mode w/ Chrome.  And how the H browser icon works

LMS007:
Screen Shot 2019-06-19 at 2 15 13 PM

dark theme

LMS007:
@hmstepanek  I dug into this a little more and we may wait 10 *more* years to see this issue resolved...
https://bugs.chromium.org/p/chromium/issues/detail?id=29683

The current docs only allow for png only.

- 16 looks to be the favicon
- 48 is the one you see in the chrome:extension page 
- 128 is from the store itself.

https://developers.chrome.com/extensions/manifest/icons

I went ahead and tried the .svg for kicks but it fails so I recommend we push this. We can use SVG for our internal icons at least.

dwhly:
> dark theme

I was thinking more the browser extension icon in the browser chrome.

hmstepanek:
Instructions for releasing a new version of the chrome extension are [here](https://github.com/hypothesis/playbook/blob/master/docs/chrome-extension.md#releasing-a-new-version-of-the-extension). There is also an internal stack overflow q/a [here](https://stackoverflow.com/c/hypothesis/questions/167).

PR #171, opened 2019-06-19
Add note about OAuthClientID to docs

LMS007:
Docs did not detail this step which is critical for local development when hosting your own instance of h.


hmstepanek:
Instructions for releasing a new version of the chrome extension are [here](https://github.com/hypothesis/playbook/blob/master/docs/chrome-extension.md#releasing-a-new-version-of-the-extension). There is also an internal stack overflow q/a [here](https://stackoverflow.com/c/hypothesis/questions/167).

PR #213, opened 2019-09-02
Fix detection of whether a tab is using Chrome's PDF viewer

### LINKED ISSUE #212: Hypothesis fails to detect PDF in Chrome
Tested by me with Chrome v78.0.3895.5 and @dwhly with Chrome 76.0.3809.132 on macOS.

**Steps to reproduce**

1. Install H Chrome extension
2. Go to http://researchoninnovation.org/patent.pdf
3. Click H icon to activate extension

**Expected result:** H loads the PDF in its custom PDF.js viewer, with the Hypothesis sidebar active
**Actual result:** The H sidebar appears in the native PDF viewer, as if the page were an HTML page. Since this is not our custom PDF.js viewer, annotation doesn't work.

Slack thread: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1567444151094900

**Technical notes**

Using Chrome dev tools with the extension's background page I've narrowed the issue down to the tab content type detection returning the wrong result ("HTML" instead of "PDF"). The browser extension detects the content type by running a script (the [`detectContentType`](https://github.com/hypothesis/browser-extension/blob/a95409f8422297592871af9c42a6bc3974528f94/src/common/detect-content-type.js#L29) function) in the context of the tab's top-level HTML document. 

In the Chrome context, it looks for an element matching the selector `embed[type="application/pdf"][name="plugin"]`.

When I looked at the actual HTML of the top-level frame in the PDF tab just now, the `embed` is present, but the `embed` does not have a `name` attribute:

Screenshot 2019-09-02 23 15 50

I don't yet know the last Chrome version where this worked. My guess at the moment is that a recent change removed the `name` attribute.

robertknight:
Chrome recently made changes to how the native PDF viewer is loaded [1],
which slightly changes the HTML of the top-level document in tabs that
are using this viewer. Specifically, the `<embed>` element no longer has a
`name` attribute. This broke the browser extension's detection of
whether the current document is a PDF by checking whether Chrome is
using its native PDF viewer to show it [2].

Resolve the issue by adjusting the CSS selector slightly.

To test, verify that activating the extension on a PDF uses the bundled
PDF.js viewer both when the "MimeHandlerViewer in cross-process frame"
feature flag in Chrome is turned on (at chrome://flags) and when it is
turned off.

Fixes #212

[1] https://bugs.chromium.org/p/chromium/issues/detail?id=659750

[2] nb. We use this method rather than relying only on the file
    extension because not all PDF URLs have a path with a `.pdf` suffix,
    and AFAIK there is no way to get the MIME type of an already-loaded
    document.

### DIFF
--- a/src/common/detect-content-type.js
+++ b/src/common/detect-content-type.js
@@ -22,12 +22,10 @@ function detectContentType(document_) {
     // instantiates the native PDF renderer.
     //
     // The selector below matches the <embed> tag in the top-level document. To
-    // see this document, open the inspector on a PDF viewer tab with
-    // Ctrl+Shift+I rather than right-clicking on the viewport and selecting the
-    // 'Inspect' option which will instead show the _inner_ document.
-    if (
-      document_.querySelector('embed[type="application/pdf"][name="plugin"]')
-    ) {
+    // see this document, open the developer tools from Chrome's menu rather
+    // than right-clicking on the viewport and selecting the 'Inspect' option
+    // which will instead show the _inner_ document.
+    if (document_.querySelector('embed[type="application/pdf"]')) {
       return { type: 'PDF' };
     }
     return null;
--- a/tests/common/detect-content-type-test.js
+++ b/tests/common/detect-content-type-test.js
@@ -19,7 +19,7 @@ describe('detectContentType()', function() {
   });
 
   it('returns "PDF" if Google Chrome PDF viewer is present', function() {
-    el.innerHTML = '<embed name="plugin" type="application/pdf"></embed>';
+    el.innerHTML = '<embed type="application/pdf"></embed>';
     assert.deepEqual(detectContentType(), { type: 'PDF' });
   });
 


src/common/detect-content-type.js
```diff
@@ -22,12 +22,10 @@ function detectContentType(document_) {
     // instantiates the native PDF renderer.
     //
     // The selector below matches the <embed> tag in the top-level document. To
-    // see this document, open the inspector on a PDF viewer tab with
-    // Ctrl+Shift+I rather than right-clicking on the viewport and selecting the
-    // 'Inspect' option which will instead show the _inner_ document.
-    if (
-      document_.querySelector('embed[type="application/pdf"][name="plugin"]')
-    ) {
+    // see this document, open the developer tools from Chrome's menu rather
```
robertknight: I changed the comment here to be more platform-independent because the previous instructions don't work on macOS. I was using Linux at the time I wrote them.

PR #215, opened 2019-09-03
Bump minimum Firefox version

robertknight:
Firefox's extension validation service is producing many warnings
complaining about APIs that are unsupported in Firefox v48 [1]. Since the FF
extension doesn't have active users at the moment, there is no harm in
bumping the minimum version to the current Firefox ESR release (v60).

[1] See eg. https://addons.mozilla.org/en-US/developers/upload/5efc192fadf74767bfe9735e006e3f0d

### DIFF
--- a/src/chrome/manifest.json.mustache
+++ b/src/chrome/manifest.json.mustache
@@ -18,9 +18,7 @@
   {{#browserIsFirefox}}
   "applications": {
     "gecko": {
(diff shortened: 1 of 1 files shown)


robertknight:
I think this is a pretty uncontroversial change since we're not actively shipping a Firefox extension to users yet. Therefore I'll go ahead and merge it.

PR #254, opened 2019-11-29
Update minimum browser versions to newer releases and remove polyfills

robertknight:
Avoid rewriting JavaScript or polyfilling APIs for very old versions of
Chrome and Firefox. This allows us to remove a couple of dependencies
(core-js, whatwg-fetch).

 - Set the supported Firefox version to the current long-term support
   (ESR) release. Since the Firefox extension is not published yet, we
   can safely require a fairly recent release
 - Set the supported Chrome version to a release from ~2 years ago.
   The choice is, roughly, the oldest release that natively supports
   the vast majority of modern JavaScript language/APIs we are likely to use


LMS007:
approved once conflicts resolved

robertknight:
I have resolved the conflicts so will merge per the previous comment.

PR #262, opened 2019-12-06
Remove an obsolete class from config script tag

robertknight:
The related client change was deployed a long time ago.


robertknight:
Since this is a trivial change and the corresponding change has already deployed in h, I'm going to merge this (see source of https://hypothes.is/app.html).

PR #282, opened 2020-01-23
Convert background page to ES modules

robertknight:
For consistency with our other projects, modernize the background page sources, which constitute most of the extensions' code, to use ES modules. There are a few other single-file scripts which are used in different contexts and have been left alone as they don't import anything.

This mostly uses the same `convert-to-es-modules` conversion script from the hypothesis/frontend-toolkit repo that was used for the client. I had to apply some of the same fixes to the codebase (eg. hoisting imports to the top level) that were also applied in the client.

### DIFF
--- a/src/background/.eslintrc
+++ b/src/background/.eslintrc
@@ -1,0 +1,5 @@
+{
+  "parserOptions": {
+    "sourceType": "module"
+  }
+}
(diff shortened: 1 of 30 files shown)


robertknight:
Thanks Lyza. I'll do another round of testing of this and get it deployed on Monday morning.

PR #298, opened 2020-02-19
Regenerate lockfile to update dependencies

robertknight:
This PR is the result of updating all non-top-level dependencies to
current compatible versions by running:

```
rm package-lock.json
npm install
```

...and committing the newly-generated package-lock.json.

This fixes a bunch of compile errors when running `npm install` due to
an outated version of fsevents.

Auditing every updated package manually is not really viable, so we're relying on `npm audit` and GitHub's security checks.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1487,9 +1487,9 @@
       }
     },
     "babel-plugin-mockable-imports": {
-      "version": "1.7.0",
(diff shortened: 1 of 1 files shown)


robertknight:
I'm going to go ahead and get this merged since, as I mentioned, auditing every detail of the change is not really practical. I created this PR mostly as information on one way to refresh indirect dependencies.

Issue #247, opened 2019-11-20
Automate the extension release process


ajpeddakotla:
### Problem
Currently our browser extension release process doesn't use the CI/CD Jenkins setup. Releasing the browser extension is therefore a more manual process, which can mean that not all team members can release the extension, or feel comfortable releasing the extension. 

### What's Needed
After talking with @robertknight, we agreed that we should move the extension release process into the current CI/CD process that we're already using for other apps. This will help automate the process so that everyone can release the extension as needed. 

### Sub-tasks

- [x] Add Jenkins project to build, test and deploy the browser extension (https://github.com/hypothesis/browser-extension/pull/300, https://github.com/hypothesis/browser-extension/pull/301)
- [x] Add method to create a new browser extension version with the latest Hypothesis client (https://github.com/hypothesis/browser-extension/pull/307, [Slack thread](https://hypothes-is.slack.com/archives/C4K6M7P5E/p1582624801011700))
- [x] Trigger an update of the QA browser extension when a new Hypothesis client version is released to production. Production releases of the browser extension will still require manual intervention (https://github.com/hypothesis/client/pull/1877)
- [x] Update [documentation in playbook repository](https://github.com/hypothesis/playbook/blob/master/docs/chrome-extension.md), SO4T to reflect extension release process changes (https://github.com/hypothesis/playbook/pull/270)
- [x] Figure out why the process of uploading, signing and then downloading the Firefox extension using `web-ext` is taking such a long time (see [comment](https://github.com/hypothesis/browser-extension/issues/247#issuecomment-596398632))
- [x] Make a decision about whether to keep or remove the Travis build for the time being (nb. The main affordance Travis provides is that it is accessible to non-Hypothesis developers, whereas Jenkins is not) (@robertknight - I'm proposing to keep this for now. We can always revisit later. See https://github.com/hypothesis/browser-extension/pull/311)

### Follow-up work

(Check items when either done or filed as an issue and scheduled for future)

- [x] Rename the "stage" extension to "QA" everywhere in the browser-extension repo, as it is confusing that we use the term "QA" for all other projects but "stage" for the browser extension pre-prod release in some places (https://github.com/hypothesis/browser-extension/pull/309)
[...]

PR #307, opened 2020-02-24
Add Jenkins build parameter to update Hypothesis client

robertknight:
Add a `BUILD_TYPE` parameter that controls whether Jenkins does a
regular build or updates the Hypothesis client and pushes a new tag.

When building with the "update-hypothesis-client" build type, the Test +
Build steps are replaced with a step that installs the latest version of
the "hypothesis" npm dependency and then creates a new npm version, as
well as creating a new tag. The changes are pushed back to the source
branch, which will always be "master", which will in turn kick off
another Jenkins build which will actually test and deploy the new
Hypothesis client release.

Remove the `tools/update-client` script as this is a manual tool for
updating the Hypothesis client which is superceded by the Jenkins task
to do this.

----

To initiate an update of the Hypothesis client in the browser extension, go to the browser-extension project in Jenkins, click "Build with Parameters" and select the "update-hypothesis-client" build type.

The screenshot below shows me doing this on a test branch, but usually you would do it from the "master" branch:

Screenshot 2020-02-24 17 38 50


robertknight:
There is a significant limitation to the current approach:

The Jenkins build that updates the Hypothesis client is not the one that actually deploys that change to production. Instead that happens as a result of a subsequent build that is triggered when the `master` branch is updated. This means that any upstream tasks (eg. a client build) waiting on the results of an `update-hypothesis-client` build will indicate success when the Hypothesis client version has been _updated in the repository_ rather than _when it has been deployed_.

An alternate solution might be to make the `update-hypothesis-client` build deploy all the way to QA, but have Jenkins ignore the subsequent build that is triggered as a result of updating the branch.

package.json
```diff
@@ -39,8 +39,6 @@
     "web-ext": "^4.1.0"
   },
   "scripts": {
-    "postversion": "git push https://github.com/hypothesis/browser-extension.git master:master --follow-tags",
-    "preversion": "npm run test",
```
robertknight: These steps were removed because the `Jenkinsfile` now handles creating the tag and pushing the version update commit back to the origin, as well as any pre-version steps.

tools/update-client
```diff
@@ -1,38 +0,0 @@
-#!/bin/sh
```
robertknight: This script was removed because updating the client is now done via a Jenkins build with the `BUILD_TYPE` parameter set to `update-hypothesis-client`.

robertknight:
I'm going to get this deployed so I can do a test run to actually do a Hypothesis client update entirely on Jenkins.

robertknight:
This was part of https://github.com/hypothesis/browser-extension/issues/247. See [this Slack thread](https://hypothes-is.slack.com/archives/C4K6M7P5E/p1582624801011700) for notes on how to use it as a developer.

PR #312, opened 2020-02-26
Update comments in chrome-webstore-refresh-token script

robertknight:
Update comments to reflect the fact that the resulting token needs to
be used in both Jenkins and Travis.


robertknight:
I'm going to go ahead and merge this as it is trivial.

PR #317, opened 2020-03-09
Upload QA and prod Firefox extensions in parallel

robertknight:
Attempt to speed up the Firefox extension signing process by uploading
both extensions and waiting for signing to complete in parallel. To make
the deployment log output easier to read, the log output from each upload is
buffered into a file and then output at once after each upload
completes.

Also make the log output easier to read by filtering out progress
messages from `webext sign`, since the tool does not follow the usual
convention of suppressing progress bars when the output is a not an
interactive terminal.

On a maintenance note, I think the logic in the `tools/deploy` script is getting to the limit of what I would comfortably put in a shell script before switching to a "proper" programming language. If we find a need to make more additions, I would probably translate this to a Node.js script first.

### DIFF
--- a/tools/deploy
+++ b/tools/deploy
@@ -38,28 +38,72 @@ upload_chrome_prod_ext() {
     --refresh-token $CHROME_WEBSTORE_REFRESH_TOKEN
 }
 
-sign_firefox_ext() {
+# Upload a Firefox extension to addons.mozilla.org for automated validation and
+# signing. Write log output to `ext_source_dir/log.txt`.
+#
+# Arguments:
+#   ext_id - ID of the Firefox extension
+#   ext_source_dir - Directory containing source files for the extension.
+# Returns:
+#   0 if extension was successfully signed and downloaded to `ext_source_dir`.
+run_webext_sign() {
+  ext_id=$1
+  ext_source_dir=$2
+
+  # Make pipeline below fail if `webext` fails.
+  set -o pipefail
+
+  # Run `web-ext sign` and filter out progress message spam, as webext is not
+  # smart enough to do this itself if not connected to a TTY.
+  node_modules/.bin/web-ext sign \
+    --api-key $FIREFOX_AMO_KEY \
+    --api-secret $FIREFOX_AMO_SECRET \
+    --id "$ext_id" \
+    --source-dir $ext_source_dir \
+    --artifacts-dir $ext_source_dir \
+    | grep -E -v "(Signing|Validating) add-on \[" \
+    | grep -E -v "Downloading signed files:" \
+    > $ext_source_dir/log.txt
+}
+
+sign_firefox_exts() {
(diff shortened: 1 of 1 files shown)


tools/deploy
```diff
@@ -38,28 +38,72 @@ upload_chrome_prod_ext() {
...
 
-sign_firefox_ext() {
+# Upload a Firefox extension to addons.mozilla.org for automated validation and
+# signing. Write log output to `ext_source_dir/log.txt`.
+#
+# Arguments:
+#   ext_id - ID of the Firefox extension
+#   ext_source_dir - Directory containing source files for the extension.
+# Returns:
+#   0 if extension was successfully signed and downloaded to `ext_source_dir`.
```
robertknight: nb. The syntax for documenting arguments and return types comes from Google's [shell script style guide](https://google.github.io/styleguide/shellguide.html).

tools/deploy
```diff
@@ -38,28 +38,72 @@ upload_chrome_prod_ext() {
...
-    --api-key $FIREFOX_AMO_KEY \
-    --api-secret $FIREFOX_AMO_SECRET \
-    --id "$FIREFOX_PROD_EXT_ID" \
-    --source-dir dist/firefox-prod \
-    --artifacts-dir dist/firefox-prod
+  # Upload the Firefox extensions and wait for signing to complete.
+  # This may take several minutes for each extension, so we do this for the
+  # QA and prod extensions in parallel.
+  run_webext_sign "$FIREFOX_QA_EXT_ID" dist/firefox-qa &
+  qa_sign_pid=$!
```
robertknight: `$!` is the process ID of the most recently launched command.

tools/deploy
```diff
@@ -38,28 +38,72 @@ upload_chrome_prod_ext() {
...
+#   ext_id - ID of the Firefox extension
+#   ext_source_dir - Directory containing source files for the extension.
+# Returns:
+#   0 if extension was successfully signed and downloaded to `ext_source_dir`.
+run_webext_sign() {
+  ext_id=$1
+  ext_source_dir=$2
+
+  # Make pipeline below fail if `webext` fails.
+  set -o pipefail
```
robertknight: Without this, `|`-separated pipeline's exit status is the exit status of the final command (ie. the `grep` calls below).

tools/deploy
robertknight: Probably not in _this_ script as it would be more of an organization-wide convention.

robertknight:
> What is the ultimate state of the log-output file if the upload processes hang or fail midway through?

Good question. I don't know for certain. I think that will depend on whether output is buffered or not by the process that hangs.

robertknight:
[Success:](https://jenkins.hypothes.is/job/browser-extension/job/master/38/console)

Screenshot 2020-03-12 15 15 47

Issue #320, opened 2020-03-18
"No tab with id" error when log in


diegodlh:
During log in, when the Log in button is pressed in the log-in popup, the log in is successful and the popup is closed, the browser extension complains about missing tabs: "No tab with id: XX" errors in the console log.
Apparently only the error is shown and functionality does not seem to be affected.
I think that the missing tab may be related to the log-in popup, which closes after successful login.

PR #321, opened 2020-03-18
Invalid tab

diegodlh:
These changes fix bug #320. They make sure the tab still exists before updating its browser action or (un)installing the sidebar on it.

### DIFF
--- a/src/background/browser-action.js
+++ b/src/background/browser-action.js
@@ -97,10 +97,14 @@ export default function BrowserAction(chromeBrowserAction) {
       }
     }
 
-    chromeBrowserAction.setBadgeText({ tabId: tabId, text: badgeText }, function() { if (chrome.runtime.lastError) {} });
-    chromeBrowserAction.setIcon({ tabId: tabId, path: activeIcon }, function() { if (chrome.runtime.lastError) {} });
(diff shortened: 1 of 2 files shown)


diegodlh:
Karma test fails because the headless chrome (which is used for testing) does not have a "chrome" object, which I use in my commits. I am not familiar with Karma at all. Can you think of any workaround?

robertknight:
The code change looks like it may be papering over a problem that happened earlier. Can you look into where the invalid tab ID is coming from in the first place, that is being passed to these functions?

diegodlh:
Hi, Robert. I think this is what is happening:
1. The login popup is updated, so the onUpdated event is fired.
2. The onTabUpdated function is run, with tabId the id number of the login popup tab.
3. Then, state.setState is run, again with tabId the id number of the login popup tab.
4. In setState, the tabId is sent to the onchange function (onTabStateChange).
5. In the onTabStateChange function, the browserAction.update(tabId, current) and updateTabDocument(tab) functions are run, but at this point the login window has been closed and the error occurs.
Maybe we could check if the tab still exists just once, in the onTabStateChange function. I'm sending new commits which do that.

Issue #316, opened 2020-03-03
Firefox extension will not install


jboeke:
I am receiving the following when trying to install my Firefox extension build.

![image](https://user-images.githubusercontent.com/547704/75797376-4ecc9600-5d32-11ea-8db4-915821088c4d.png)

## Steps to Reproduce:

1. Clone [repo](https://github.com/hypothesis/browser-extension) at commit ea8fa71015616587e9e0cb17fd0f552d05c9094f.

2. On Win 10, using MINGW64 shell provided by `git bash` (with administrative permissions), use Chocolatey to install make tools via the command `choco install make`

3. `git checkout -b my-branch` in order to make a new local branch.

4. Delete all files in `build` folder.

5. Commit changes locally on `my-branch`.

6. Make with command `make SETTINGS_FILE=settings/firefox-prod.json`. Output appears to be successful:

```
$ make SETTINGS_FILE=settings/firefox-prod.json
node_modules/.bin/browserify -t babelify -d src/background/index.js > build/extension.bundle.js.tmp
cat build/extension.bundle.js.tmp | node_modules/.bin/exorcist build/extension.bundle.js.map >build/extension.bundle.js
node_modules/.bin/mustache build/.settings.json src/manifest.json.mustache > build/manifest.json
cp -R node_modules/hypothesis/build/* build/client/build
rm build/client/build/manifest.json
tools/template-context-app.js build/.settings.json | node_modules/.bin/mustache - src/sidebar-app.html.mustache >build/client/app.html
tools/template-context-settings.js build/.settings.json | node_modules/.bin/mustache - src/settings-data.js.mustache >build/settings-data.js
cp src/unload-client.js build/unload-client.js
cp src/pdfjs-init.js build/pdfjs-init.js
cp -R src/vendor/pdfjs build/pdfjs
cp -R src/help/* build/help
cp -R src/images/* build/images
cp -R src/options/* build/options
```
7. Zip all files in build directory into `build.zip`

8. Attempt install of the file from the previous step on `about:addons` page.

![image](https://user-images.githubusercontent.com/547704/75798124-5b9db980-5d33-11ea-83e9-51277afe68d9.png)

I'm happy to include the zip file here, but did not want to accidentally pass on any sensitive data inside it.

robertknight:
Hello,

[...]

Issue #310, opened 2020-02-26
Publish Firefox extension on addons.mozilla.org


QJKX:
Creating an issue so I can be notified when it's available 🙂.

robertknight:
Hello,

Our current plan is to host the Firefox extension ourselves (ie. make it available from https://web.hypothes.is) rather than list it on https://addons.mozilla.org, although it will still go through Mozilla's automated checking processes. The reason is because we make hundreds of small updates each year and addons listed on addons.mozilla.org have to go through a manual (rather than automated) review process, making it difficult to ensure users get updates quickly.

QJKX:
Thanks for the reply.
That's pragmatic but disappointing.

I suppose fixing that needs support from Mozilla's end (maybe through the "recommended extension" program?)
https://support.mozilla.org/en-US/kb/recommended-extensions-program#w_may-i-suggest-extensions-to-become-recommended

anavarre:
It's great you're considering a Firefox extension. An ETA would be really nice to see so expectations are set. If you need beta testers, it'd be great to know how to participate as well.

> The reason is because we make hundreds of small updates each year and addons listed on addons.mozilla.org have to go through a manual (rather than automated) review process, making it difficult to ensure users get updates quickly.

While I understand the reasoning, I'd argue it's better to have something than nothing, so you could continue working at a fast pace and offer the extension via the official website but also publish on addons.mozilla.org as well. This could be on a quarterly basis for instance, and it'd increase discoverability of the service, bug reports / feature requests and overall popularity.

robertknight:
>  An ETA would be really nice to see so expectations are set. If you need beta testers, it'd be great to know how to participate as well.

Thanks for the offer to help with testing. We have no ETA at present.

ZiadJ:
So, if I got it right, Firefox support is not being considered as a priority for now. The only reason why I haven't recommended the Chrome plugin to others so far is because it is not cross-browser compatible so remains inaccessible by a good portion of the population. Firefox is the only major open source browser out there and those using it cannot be ignored if we want true freedom of information.

I'm wondering what the actual priorities are then if it's not to provide a seamless support across major browsers in a non-fragmented way. Again, I can't use nor can I recommend Hypothesis in its currently fragmented state so I'm really curious to know what those other priorities are that would make not worth an ETA. So my question is: Are the development requirements too high somehow for it to be worth it and, if so, what exactly is to be considered as more worthy for now?

robertknight:
[...]

PR #341, opened 2020-05-03
Add Identity API support for login

diegodlh:
This adds [Identity API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/identity) support to allow login with Firefox extension, as discussed [here](https://github.com/hypothesis/browser-extension/issues/316#issuecomment-602286222).

API support introduced here can be used by the client as proposed in a separate [pull request](https://github.com/hypothesis/client/pull/2107).

To have this working with the production h server, the corresponding OAuth client has to be configured in the h server with the redirect url returned by the extension's [chrome.identity.getRedirectURL()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/identity/getRedirectURL). For example, in Firefox this is of the form https://<id-derived-from-app-id>.extensions.allizom.org/.

To simplify testing in development, I changed src/manifest.json.mustache to allow explicitly setting the [add-on's ID](https://extensionworkshop.com/documentation/develop/extensions-and-the-add-on-id/) in settings/firefox-dev.json, just like it is possible to set a key in settings/chrome-dev.json.

What does not work:
1) I haven't found a way to change the size of the auth popup window. It seems to be [unsupported](https://stackoverflow.com/questions/18958944/setting-popup-dimensions-in-chrome-launchwebauthflow).
2) The cancel button in the login page that opens in the popup (hypothes.is/login?...) can't close the popup. It fails with "Scripts may not close windows that were not opened by script".
One possible solution could be to add a parameter to the authUrl passed to launchWebAuthFlow so that the h server knows that we are using this auth method and either (i) remove then 'cancel' button, or (ii) make the button send a message when clicked asking the extension to close the window instead. If you think this makes sense, I could try and make the necessary changes to the 'h' repo.

I tried to test this in Firefox for Android, but it doesn't work because it doesn't support the Identity API.
I also tried to test it in [Firefox Preview](https://support.mozilla.org/en-US/products/firefox-preview), the upcoming version of Firefox for Android, and it installed, but I couldn't try it because I couldn't easily find how to have the development h server listen to connections from my LAN (i.e., listen on 0.0.0.0:5000 instead of 127.0.0.1:5000). Can you help me figure this out?


diegodlh:
Dear @robertknight, do you think you'll have the time to consider this pull request? I have been [suggested](https://github.com/hypothesis/browser-extension/issues/310#issuecomment-637715653) to pack an unofficial Firefox extension and I like the idea, but I wanted to make sure that you won't be able consider my pull requests before proceeding with this separate effort. Thanks!

diegodlh:
Just for the record, this has been implemented in an [unofficial extension](https://github.com/diegodlh/browser-extension/releases/latest). See discussion [here](https://github.com/hypothesis/browser-extension/issues/310#issuecomment-637715653).

PR #413, opened 2020-10-22
Use yarn instead of npm

esanzgar:
For all other hypothesis projects we use `yarn`.

This would make the use of `yarn` consistent among all the frontend projects.


Jenkinsfile
```diff
@@ -40,7 +40,7 @@ node {
 
     stage('Setup') {
       nodeEnv.inside("-e HOME=${workspace}") {
-        sh "npm ci"
+        sh "yarn install"
```
robertknight, line 43: I found a remaining reference to `npm install` in `docs/building.md`.

Dockerfile
```diff
@@ -3,14 +3,15 @@
...
 # Alpine is updated every 6 months so all packages are pretty recent.
-FROM alpine:3.10
+FROM alpine:3.12
 
 RUN apk update && apk add --no-cache \
   chromium \
   git \
   make \
   nodejs \
   npm \
```
robertknight, line 13: I'm not sure if we still need npm here. I note that it still exists in the client's Dockerfile.

robertknight (approved):
> For all other hypothesis projects we use yarn.

That's not quite true. For the biggest frontend projects (client, lms) we do. The h repo still uses npm and there are some smaller repositories where npm vs yarn doesn't make much difference.

We agreed on Slack though that using yarn makes sense here as the lockfile format is easier to diff, the output is less noisy and there are some other minor quality-of-life benefits.

I found one remaining reference to `npm install` in `docs/building.md` which needs updating.

Dockerfile
esanzgar: I confirm `npm` it is not needed.

In the client I see a use of  `npm publish` command, which could be modified to `yarn publish`

PR #414, opened 2020-10-22
Update babel

esanzgar:
From version 6 to @babel/core version 7.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,10 +1,13 @@
 {
   "presets": [
-    ["env", {
-      "targets": {
-        "firefox": "68",
-        "chrome": "64"
+    [
+      "@babel/preset-env",
+      {
+        "targets": {
+          "chrome": "64",
+          "firefox": "68"
+        }
       }
-    }]
+    ]
   ]
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 build/
 dist/
 node_modules/
+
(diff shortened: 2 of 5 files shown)


robertknight (approved):
LGTM. I confirmed that async/await syntax is handled as expected now (ie. left as-is and not compiled).

I used the "Edit" button in the top-right corner of the page to change the _base_ for this PR from `master` to the previous PR in this series (`use-yarn`). That makes GitHub show only the diff compared to that branch. When the `use-yarn` branch is merged to `master`, this PR will be automatically updated to target `master`.

esanzgar:
While debugging a failing test I found the following line:
`at asyncGeneratorStep (/var/folders/w2/6l_zlgwn1zx6whm0hzym363w0000gn/T/cf422efb520d18e976103b0ba434963a.browserify.js:6345:103)`

Isn't this indicative that babel wraps async/await in generators during testing. If so, why?

robertknight:
> Isn't this indicative that babel wraps async/await in generators during testing. If so, why?

Indeed it seems so. Looking in `karma.config.js` it appears there is an override of the `.babelrc` configuration which we'll need to remove: https://github.com/hypothesis/browser-extension/blob/383d119319def2a5a1a98ea902797748d3e615cb/tests/karma.config.js#L68

I'm guessing this dates back to when PhantomJS was used instead of Chrome to run the tests, since PhantomJS doesn't support any ES6 features.

PR #415, opened 2020-10-23
Use same compilation settings in tests and bundle build

robertknight:
Remove the `presets` override in `karma.config.js` and resolve an issue
that prevented configuration in the `.babelrc` file being used for
source files located outside of the `tests/` directory.

Using the same compilation settings ensures that the test environment
better matches the actual build. It also fixes a problem with using
async/await syntax and other modern JS syntax in tests / code being
tested.

Slack debugging thread: https://hypothes-is.slack.com/archives/D01AJ11PUER/p1603459022001600

### DIFF
--- a/tests/karma.config.js
+++ b/tests/karma.config.js
@@ -63,15 +63,19 @@ module.exports = function (config) {
 
     browserify: {
       debug: true,
-      configure: function (bundle) {
-        bundle.transform('babelify', {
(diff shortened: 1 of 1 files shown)


esanzgar (approved):
I confirm it works as expected:
- no errors with exports and imports
- no use of generators wrappers around async/awaits

PR #417, opened 2020-10-27
eslint update

esanzgar:
This PR consist on two commits:

1. First commit is to adopt the latest version of eslint-config-hypothesis. I had to install a few supporting eslint plugin libraries.
2. To pass the linting I had to convert `var` to `let` or `const`.


robertknight (approved):
Looks good to me.

PR #421, opened 2020-10-27
Correct `yarn version` syntax

robertknight:
When `npm` was replaced with `yarn`, the syntax for specifying the new
version needed to be updated.

See https://classic.yarnpkg.com/en/docs/cli/version/.


esanzgar (approved):
I confirm the fix is correct.

PR #422, opened 2020-10-29
Reduce badge request at parent level

esanzgar:
The `tab.onUpdated` event is fired many times during the loading of a page. The larger the document the more times the event is triggered.

By moving the method that request badges to a place that is only reached when the page is fully loaded, we avoid unnecessary fetch request to the badge end point.

### DIFF
--- a/src/background/hypothesis-chrome-extension.js
+++ b/src/background/hypothesis-chrome-extension.js
@@ -153,14 +153,13 @@ export default function HypothesisChromeExtension(dependencies) {
     return activeState;
   }
 
-  function resetTabState(tabId, url) {
+  function resetTabState(tabId) {
     state.setState(tabId, {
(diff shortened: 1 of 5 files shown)


robertknight:
> By moving the method that request badges to a place that is only reached when the page is fully loaded, we avoid unnecessary fetch request to the badge end point.

With the caveat that I haven't tested this PR yet, I'm fairly sure this was explored previously. A problem we ran into is that on ad-heavy websites (eg. many major news websites) it takes an extremely long time for the page to completely load, but the actual article content is ready much sooner.

esanzgar:
@robertknight, you are correct: moving the badge request to be invoked when the document is fully loaded will delay the request for ad-heavy pages.

Unfortunately, this is not equivalent to `DOMContentLoaded`.

PR #423, opened 2020-10-29
Reset the badge count to 0 in case of badge fetch failures, not cancellation

esanzgar:
If errors arise during the fetching of badge requests it reset the count to 0.
    
In case of cancellations, do not reset the count to 0.

### DIFF
--- a/src/background/errors.js
+++ b/src/background/errors.js
@@ -12,6 +12,8 @@ export class BlockedSiteError extends ExtensionError {}
 
 export class AlreadyInjectedError extends ExtensionError {}
 
+export class RequestCanceledError extends Error {}
+
 /**
  * Returns true if `err` is a recognized 'expected' error.
  */
--- a/src/background/tab-state.js
+++ b/src/background/tab-state.js
@@ -1,4 +1,5 @@
 import isShallowEqual from 'is-equal-shallow';
+import { RequestCanceledError } from './errors';
 
 import * as uriInfo from './uri-info';
 
@@ -40,7 +41,7 @@ const DEFAULT_STATE = {
  * initialState - An Object of tabId/state keys. Used when loading state
  *   from a persisted store such as localStorage. This will be merged with
  *   the default state for a tab.
- * onchange     - A function that recieves onchange(tabId, current).
+ * onchange     - A function that receives onchange(tabId, current).
  */
 export default function TabState(initialState, onchange) {
   const self = this;
@@ -48,7 +49,7 @@ export default function TabState(initialState, onchange) {
 
   /**
    * @typedef BadgeRequest - this object is used to cancel a pending badge request
-   * @prop {function} cancel - cancellation function to abort pending request
+   * @prop {function} cancel - cancelation function to abort pending request
    * @prop {number} waitMs - the current waiting time after which the request will be invoked
    */
 
@@ -168,7 +169,7 @@ export default function TabState(initialState, onchange) {
 
     pendingRequest?.cancel();
 
-    const debouncedFetch = new Promise(resolve => {
+    const debouncedFetch = new Promise((resolve, reject) => {
       const timerId = setTimeout(async () => {
         let count;
         try {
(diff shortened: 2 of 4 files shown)


src/background/tab-state.js
```diff
@@ -183,14 +183,22 @@ export default function TabState(initialState, onchange) {
       pendingAnnotationCountRequests.set(tabId, {
         cancel: () => {
           clearTimeout(timerId);
-          resolve(0);
+          reject('Badge request cancelled');
```
robertknight, line 186: As a convention we try to ensure that promises are always rejected with `Error` instances (or `Error`-like objects). This is convenient when catching the error as you can make some assumptions about the type of the exception value (eg. that it always has a `message` property).

We have a set of existing error subclasses defined in `errors.js`, so I would recommend extending that.

robertknight (changes requested):
Hi @esanzgar,

I had a recommendation about changing the type of the value used to reject the promise from a string to an `Error` subclass. Please see the comments for details.

src/background/tab-state.js
esanzgar: I agree.

I added a new error.

esanzgar:
I created a different Error class for these cancellations.

src/background/errors.js
```diff
@@ -12,6 +12,8 @@ export class BlockedSiteError extends ExtensionError {}
 
 export class AlreadyInjectedError extends ExtensionError {}
 
+export class BadgeRequestError extends Error {}
```
robertknight, line 15: The name here suggests that this could refer to _any_ error that happened when fetching the badge count, but from the implementation it looks like this is only used if the badge request is canceled? If so I would suggest to rename this to reflect that - the exception could either specifically reflect the fact that the badge request was canceled (maybe `BadgeRequestCanceledError`?) or perhaps a more generic exception to reflect that an operation was explicitly canceled (`CanceledError` ?).

src/background/tab-state.js
```diff
@@ -183,14 +184,22 @@ export default function TabState(initialState, onchange) {
       pendingAnnotationCountRequests.set(tabId, {
         cancel: () => {
           clearTimeout(timerId);
-          resolve(0);
+          reject(new BadgeRequestError('Badge request cancelled'));
```
robertknight, line 187: It seems like we use a mix of spellings of "cancelled" (one vs two Ls). I would suggest we favor the single-L version as various sources say that is more common in American English and is also what various web specifications use.

robertknight (changes requested):
The exception name `BadgeRequestError` sounds like it refers to any kind of error during a badge request. Since we only use it for cancelation, I would prefer to rename it to reflect that.

src/background/errors.js
esanzgar: I think the combination of `Cancel` and `Error` is not clear, it seems to indicate an error raised _during_ the cancelation. I would prefer `RequestError` as a more generic error.

src/background/tab-state.js
esanzgar: OK.

src/background/errors.js
```diff
@@ -12,6 +12,8 @@ export class BlockedSiteError extends ExtensionError {}
 
 export class AlreadyInjectedError extends ExtensionError {}
 
+export class RequestError extends Error {}
```
robertknight, line 15: This `RequestError` name unfortunately has the same problem that I raised previously - it suggests the error might be _any_ kind of request error, but you're only using it to specifically mean that the request was canceled. I think something like `RequestCanceledError` makes it clear that the exception was raised _because_ the request was canceled, rather than when attempting to cancel the request. As evidence for this I'd point to a number of APIs in other languages having [CancelledError](https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError), [CanceledException](https://developer.android.com/reference/android/app/PendingIntent.CanceledException) or similarly named errors for this purpose.

An alternative direction which I think would be semantically clear would be to use the same `RequestError` type for all kinds of error making the request but add a field (`reason`?) that allows distinguishing cancellation from other reasons (eg. network failure, unexpected response structure). You'd have to make a bunch of changes in other places to do this though.

esanzgar: 👍🏼

robertknight (approved):
LGTM

PR #424, opened 2020-10-29
Skip badge request if URL is the cache

esanzgar:
In order to further reduce the number of badge requests we implement a cache or badge request results. This cache is temporal and it contains key-value pairs of URLs and annotation counts.

* every entry in the cache (URL and annotation count pair) is added only after successful fetching of the annotation count.
* fragment is removed from the URL
* every entry in the cache  is stored for X amount of time. After that time the entry is deleted. This is to allow updating the annotation count in case of reloading of the document in the tab, after X amount of time. 

Currently, X is set to 3000 ms.

I feel this could lead to minor frustrations for users that are reloading the document in less than X amount of time to see changes in the annotation count.

### DIFF
--- a/src/background/tab-state.js
+++ b/src/background/tab-state.js
@@ -57,6 +57,10 @@ export default function TabState(initialState, onchange) {
   /** @type {Map<number, BadgeRequest>} */
   const pendingAnnotationCountRequests = new Map();
 
+  // This variable contains a map between a URI and the associated annotation count.
+  /** @type {Map<string, number>} */
+  const annotationCountCache = new Map();
+
   this.onchange = onchange || null;
 
   /** Replaces the H state for all tabs with the state data
@@ -160,8 +164,23 @@ export default function TabState(initialState, onchange) {
   this.updateAnnotationCount = async function (tabId, tabUrl) {
     const INITIAL_WAIT_MS = 1000;
     const MAX_WAIT_MS = 3000;
+    const CACHE_EXPIRATION_MS = 3000;
 
     const pendingRequest = pendingAnnotationCountRequests.get(tabId);
+
+    let url;
+    try {
+      url = uriInfo.uriForBadgeRequest(tabUrl);
+    } catch {
+      return;
+    }
+
+    const annotationCount = annotationCountCache.get(url);
+    if (annotationCount !== undefined) {
+      this.setState(tabId, { annotationCount });
+      return;
+    }
+
     const wait = Math.min(
       pendingRequest?.waitMs ?? INITIAL_WAIT_MS,
       MAX_WAIT_MS
@@ -171,9 +190,19 @@ export default function TabState(initialState, onchange) {
 
     const debouncedFetch = new Promise((resolve, reject) => {
       const timerId = setTimeout(async () => {
-        let count;
+        let count = annotationCountCache.get(url);
+        if (count !== undefined) {
+          resolve(count);
+          return;
+        }
+
         try {
-          count = await uriInfo.getAnnotationCount(tabUrl);
+          count = await uriInfo.fetchAnnotationCount(url);
+          annotationCountCache.set(url, count);
+          setTimeout(
+            () => annotationCountCache.delete(url),
+            CACHE_EXPIRATION_MS
+          );
         } catch {
           count = 0;
         }
--- a/src/background/uri-info.js
+++ b/src/background/uri-info.js
@@ -1,4 +1,5 @@
 import settings from './settings';
+import { BadgeUriError } from './errors';
 
 const ALLOWED_PROTOCOLS = new Set(['http:', 'https:']);
 
@@ -16,30 +17,33 @@ function encodeUriQuery(val) {
 }
 
 /**
- * Based on the request protocol and hostname decide if the URL should be sent to the "badge"
- * request endpoint.
+ * Returns a normalized version of URI for use in badge requests, or throws BadgeUrlError
+ * if badge requests cannot be made for the given URL
+ *
+ * The normalization consist on (1) adding a final '/' at the end of the URL and
+ * (2) removing the fragment from the URL. The URL fragment can be ignored as it
+ *  will result the same badge count.
+ *
+ *  In addition, this normalization facilitates the identification of unique URLs.
  *
  * @param {string} uri
- * @return {boolean} - false if the URL should not be sent to the "badge" request endpoint
+ * @return {string} - URL without fragment
+ * @throws Will throw if URL is invalid or should not be sent to the 'badge' request endpoint
  */
-function shouldQueryUri(uri) {
-  let url;
-
-  try {
-    url = new URL(uri);
-  } catch (e) {
-    return false;
-  }
+export function uriForBadgeRequest(uri) {
+  const url = new URL(uri);
 
   if (!ALLOWED_PROTOCOLS.has(url.protocol)) {
-    return false;
+    throw new BadgeUriError('Blocked protocol');
   }
 
   if (BLOCKED_HOSTNAMES.has(url.hostname)) {
-    return false;
+    throw new BadgeUriError('Blocked hostname');
   }
 
-  return true;
+  url.hash = '';
+
+  return url.toString();
 }
 
 /**
@@ -47,8 +51,9 @@ function shouldQueryUri(uri) {
  * statistics about the annotations for a given URL.
  *
  * @return {Promise<number>}
+ * @throws Will throw a variety of errors: network, json parsing, or wrong format errors.
  */
-async function query(uri) {
+export async function fetchAnnotationCount(uri) {
   const response = await fetch(
     settings.apiUrl + '/badge?uri=' + encodeUriQuery(uri),
     {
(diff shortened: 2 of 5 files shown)


src/background/tab-state.js
```diff
@@ -160,8 +164,23 @@ export default function TabState(initialState, onchange) {
   this.updateAnnotationCount = async function (tabId, tabUrl) {
     const INITIAL_WAIT_MS = 1000;
     const MAX_WAIT_MS = 3000;
+    const EXPIRATION_CACHE_MS = 3000;
```
robertknight, line 167:
```suggestion
    const CACHE_EXPIRATION_MS = 3000;
```

Suggested change of word order to be more consistent with the other constants here.

src/background/tab-state.js
```diff
@@ -160,8 +164,23 @@ export default function TabState(initialState, onchange) {
...
 
     const pendingRequest = pendingAnnotationCountRequests.get(tabId);
+
+    let url;
+    try {
+      url = uriInfo.shouldQueryUri(tabUrl);
+    } catch {
```
robertknight, line 174: This generic catch could potentially mask unexpected errors. Looking at the current implementation of `shouldQueryUri` it will currently throw either `BadgeUriError` if the URI is blocked or `TypeError` if the URI failed to parse. If `shouldQueryUri` caught the latter exception and converted it to a `BadgeUriError` then you could catch just `BadgeUriError` exceptions here and re-throw other, unexpected exceptions.

src/background/tab-state.js
```diff
@@ -171,9 +190,19 @@ export default function TabState(initialState, onchange) {
...
+          return;
+        }
+
         try {
-          count = await uriInfo.getAnnotationCount(tabUrl);
+          count = await uriInfo.fetchAnnotationCount(url);
+          annotationCountCache.set(url, count);
+          setTimeout(
+            () => annotationCountCache.delete(url),
+            EXPIRATION_CACHE_MS
```
robertknight, line 204: I tried to think of scenarios where we could end up removing the cache entry at the wrong time or have multiple timeouts for the same URL, but I think this is OK because there can only be one pending request at once.

tests/background/tab-state-test.js
```diff
@@ -136,13 +137,15 @@ describe('TabState', () => {
...
+    const EXPIRATION_CACHE_MS = 3000;
 
     beforeEach(() => {
       clock = sinon.useFakeTimers();
       getStub = sinon.stub();
       $imports.$mock({
         './uri-info': {
-          getAnnotationCount: getStub,
+          fetchAnnotationCount: getStub,
+          shouldQueryUri,
```
robertknight, line 148: Rather than calling through to `shouldQueryUri` directly here, you could mock it to directly throw a `BadgeUrlError`. This avoids coupling this test to the details of `shouldQueryUri`.

tests/background/uri-info-test.js
```diff
@@ -61,39 +33,90 @@ describe('background/uri-info', () => {
...
+        }
       });
     });
 
-    ['https://www.google.com', 'http://www.example.com'].forEach(okURI => {
-      it('sends request to API if URI does not match blocklist entries and has an allowed protocol', async () => {
-        assert.strictEqual(await uriInfo.getAnnotationCount(okURI), 1);
-        assert.equal(fetchStub.callCount, 1);
+    ['https://www.google.com/', 'http://www.example.com/'].forEach(okURI => {
+      it('sends request to API if URI does not match blocklist entries and has an allowed protocol', () => {
```
robertknight, line 47: This test description needs updating as the function being tested (`shouldQueryUri`) doesn't "send a request to the API" but instead "returns a normalized URI".

tests/background/uri-info-test.js
```diff
@@ -61,39 +33,90 @@ describe('background/uri-info', () => {
...
+        assert.strictEqual(uriInfo.shouldQueryUri(okURI), okURI);
+      });
+    });
+
+    ['https://www.google.com#1', 'https://www.google.com#2'].forEach(okURI => {
```
robertknight, line 52: I'd suggest to make one of these tests a test for the empty fragment. Currently the two test cases here are essentially the same.

src/background/uri-info.js
```diff
@@ -16,39 +17,37 @@ function encodeUriQuery(val) {
...
 
 /**
  * Queries the Hypothesis service that provides
  * statistics about the annotations for a given URL.
  *
  * @return {Promise<number>}
+ * @throws Will throw a variety of errors: network, json parsing, or wrong format errors.
  */
-async function query(uri) {
+export async function fetchAnnotationCount(uri) {
```
robertknight, line 50: `fetchAnnotationCount` is a clearer name than `query` 👍

robertknight (changes requested):
The implementation method looks good. The main feedback I have is that although the semantics and usage of `shouldQueryUri` have changed, it still has its previous name which suggests the old API. I would like to see the function name changed to reflect its new behaviour. I also had a recommendation regarding the kinds of errors that `shouldQueryUri` throws and how they are caught.

src/background/tab-state.js
esanzgar: I can't think of a scenario where you want to act differently to the `BadgeUrlError` or  the `TypeError` (from failing to parse the URL).

tests/background/tab-state-test.js
esanzgar: Good idea!

src/background/tab-state.js
esanzgar: It can only be one request pending _per tab_ at at time. There could be multiple request for the same URL for _different  tabs_.

In both cases, I can't think of any issues regarding removing the cache entry: it will always be `x`ms after the last successful request.

tests/background/tab-state-test.js
```diff
@@ -152,12 +156,107 @@ describe('TabState', () => {
...
+      const promise = tabState.updateAnnotationCount(1, 'invalidOrblocked');
+      clock.tick(INITIAL_WAIT_MS);
+
+      await promise;
+      assert.notCalled(getStub);
+      assert.equal(tabState.getState(1).annotationCount, 0);
+    });
+
+    it('updates the annotationCount (immediately) if previous URL query is still in the cache', async () => {
+      uriForBadgeRequestStub.returnsArg(0);
```
robertknight, line 174: If the same initialization is repeated for a stub in many places you can set it as a "default" behavior in the `beforeEach` block or other function that creates the stub, with something like: `uriForBadgeRequestStub = sinon.stub().returnsArg(0)`.

tests/background/tab-state-test.js
```diff
@@ -134,15 +134,19 @@ describe('TabState', () => {
...
     let getStub;
+    let uriForBadgeRequestStub;
     const INITIAL_WAIT_MS = 1000;
     const MAX_WAIT_MS = 3000;
+    const CACHE_EXPIRATION_MS = 3000;
 
     beforeEach(() => {
       clock = sinon.useFakeTimers();
       getStub = sinon.stub();
+      uriForBadgeRequestStub = sinon.stub();
```
robertknight, line 145:
```suggestion
      uriForBadgeRequestStub = sinon.stub().returnsArg(0);
```

Rather than configuring this stub's behavior in many separate tests, it looks like it would be simpler to do it here.

tests/background/tab-state-test.js
```diff
@@ -134,15 +134,19 @@ describe('TabState', () => {
...
       $imports.$mock({
         './uri-info': {
-          getAnnotationCount: getStub,
+          fetchAnnotationCount: getStub,
```
robertknight, line 148: Might be a good time to rename `getStub` to `fetchAnnotationCountStub`.

tests/background/uri-info-test.js
```diff
@@ -1,46 +1,17 @@
...
-            );
-          });
+  describe('uriForBadgeRequest', () => {
+    ['http://.com', 'https://', 'dummy.com'].forEach(badURI => {
+      it('throws for invalid URL', () => {
+        try {
+          uriInfo.uriForBadgeRequest(badURI);
+        } catch (error) {
+          assert.instanceOf(error, TypeError);
+          assert.equal(error.message, "Failed to construct 'URL': Invalid URL");
```
robertknight, line 13: The text of this message is not guaranteed across browsers or browser versions, so we may need to update this in future.

robertknight (approved):
LGTM. I had a small suggestion regarding reducing repeated initialization of a stub in the tests.

tests/background/tab-state-test.js
esanzgar: Yes, I agree. I tried but then I couldn't overwrite it later.

tests/background/tab-state-test.js
esanzgar: 👍🏼

tests/background/tab-state-test.js
esanzgar: I have tried again and it is working. Thanks for the suggestion.

tests/background/uri-info-test.js
esanzgar: You are right, I will just check for the type of error that is consistent across browsers.

PR #427, opened 2020-11-02
Update instructions to open local PDFs

esanzgar:
Instruction were outdated. Now, the page links to a evergreen KB article.

image


lyzadanger:
I like the overall tone and clarity here...
I'm wondering if the following phrasing might be a little clearer:

"If you're trying to use a PDF document stored on your computer, you'll need to change your settings to allow us to access these files."

Specifically, "it's" is confusing here (What is "it"?) and "local" is a little bit technical given the relaxed tone we're using here.

klemay:
How about:

"You may need to change your settings. Follow these steps to access PDF files stored on your computer."

? 

(I am fine with what Lyza has suggested, but I'm thinking shorter and more concise may be better!)

lyzadanger:
`You may need to change your settings. Follow these steps to access PDF files stored on your computer.`

+1

esanzgar:
image

lyzadanger (approved):

PR #428, opened 2020-11-02
Enable type checking

robertknight:
This PR adds type checking to the browser-extension project in a similar manner to how we have it set up for the client and lms projects. It also fixes the small number of existing type checking errors.

This will help catch mistakes in JSDoc documentation, incorrect API usage etc.

To run type checking locally use `yarn typecheck` or `make lint`. If you have an editor integration set up already that should also flag errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   "devDependencies": {
     "@babel/core": "^7.1.6",
     "@babel/preset-env": "^7.1.6",
+    "@types/chrome": "^0.0.126",
     "babel-plugin-mockable-imports": "^1.7.0",
     "babelify": "^10.0.0",
     "browserify": "^16.5.2",
@@ -39,11 +40,13 @@
     "query-string": "^3.0.1",
     "raven-js": "^3.27.2",
     "sinon": "^9.2.0",
+    "typescript": "^4.0.5",
     "watchify": "^3.11.1",
     "web-ext": "^5.3.0"
   },
   "scripts": {
-    "test": "karma start tests/karma.config.js --single-run"
+    "test": "karma start tests/karma.config.js --single-run",
+    "typecheck": "tsc --build src/tsconfig.json"
   },
   "dependencies": {}
 }
--- a/src/background/browser-name.js
+++ b/src/background/browser-name.js
@@ -4,6 +4,7 @@
  * @return {'chrome'|'firefox'}
  */
 export default function browserName() {
+  // @ts-expect-error - `browser` is missing from types
   if (window.browser) {
     return 'firefox';
   } else {
--- a/src/background/raven.js
+++ b/src/background/raven.js
@@ -6,7 +6,7 @@
  * version to be provided via the app's settings object.
  */
 
-import Raven from 'raven-js';
+import * as Raven from 'raven-js';
 
 /**
  * Returns the input URL if it is an HTTP URL or the filename part of the URL
@@ -69,7 +69,7 @@ export function init(config) {
 /**
  * Report an error to Sentry.
  *
- * @param {Error} error - An error object describing what went wrong
+ * @param {Object} error - An error object describing what went wrong
  * @param {string} when - A string describing the context in which
  *                        the error occurred.
  * @param {Object} [context] - A JSON-serializable object containing additional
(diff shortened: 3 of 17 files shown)


package.json
```diff
@@ -10,6 +10,7 @@
   "devDependencies": {
     "@babel/core": "^7.1.6",
     "@babel/preset-env": "^7.1.6",
+    "@types/chrome": "^0.0.126",
```
robertknight, line 13: These are community-provided types for Chrome extension APIs, from the DefinitelyTyped project. See https://www.npmjs.com/package/@types/chrome.

src/background/raven.js
```diff
@@ -69,7 +69,7 @@ export function init(config) {
 /**
  * Report an error to Sentry.
  *
- * @param {Error} error - An error object describing what went wrong
+ * @param {Object} error - An error object describing what went wrong
```
robertknight, line 72: The argument passed to this function was not always an `Error` and the code checked for this. This change just updates the documentation to reflect that.

src/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "module": "commonjs",
+    "noEmit": true,
+    "strictNullChecks": true,
+    "target": "ES2020"
+  },
+  "include": [
+    "**/*.js"
+  ],
+  "exclude": [
+    "vendor/"
```
robertknight, line 17: This excludes the local copy of PDF.js from being checked.

src/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "checkJs": true,
+    "lib": ["es2018", "dom"],
+    "jsx": "react",
+    "jsxFactory": "createElement",
+    "module": "commonjs",
+    "noEmit": true,
+    "strictNullChecks": true,
```
esanzgar, line 10: Maybe we should consider the possibility to add `"strict": true`. It checks for more than just `strictNullChecks`
https://www.typescriptlang.org/docs/handbook/compiler-options.html

Maybe we could allow for `implicitAny`. I suggest this:
```
    "strict": true,
    "noImplicitAny": false,
```

src/background/raven.js
esanzgar: What about `Error|object`? The net effect is the same but it is more informative to the developer.

It is recommended to use `object` instead of `Object`: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

src/background/browser-name.js
```diff
@@ -4,6 +4,7 @@
  * @return {'chrome'|'firefox'}
  */
 export default function browserName() {
+  // @ts-expect-error - `browser` is missing from types
```
esanzgar, line 7: Is `ts-ignore` a better option?
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#ts-ignore-or-ts-expect-error

src/help/index.js
```diff
@@ -26,6 +26,8 @@ chrome.runtime.getPlatformInfo(function (info) {
 
 const query = parseQuery(window.location.search);
 if (query.message) {
-  const errorTextEl = document.querySelector('.js-error-message');
+  const errorTextEl = /** @type {HTMLElement} */ (document.querySelector(
```
esanzgar, line 29: I tried without the casting and it doesn't complain. `textContent` is present in `Element` type, I believe.

esanzgar (approved):
Added a few minor comments.

src/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "checkJs": true,
+    "lib": ["es2018", "dom"],
+    "jsx": "react",
+    "jsxFactory": "createElement",
```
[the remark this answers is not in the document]
robertknight: Indeed. I'll remove it. If we make the Options dialog more complicated in future we might want to introduce Preact, but there is no need right now.

src/background/raven.js
robertknight: I don't actually know if it save to assume that `error` will be an object as opposed to a primitive in this context. This is a typical case of enabling TypeScript turning up existing uncertainty / ambiguity in the code. I'll leave it as `Object`/`any` for now to make it clear that it can't be assumed.

PR #429, opened 2020-11-03
Simplify custom error definitions

robertknight:
Use ES classes to simplify the definition of custom error types in the
extension.

All of the custom constructors can be removed because derived classes automatically forward constructor arguments to their base class.


esanzgar (approved):
Well done.

PR #436, opened 2020-11-11
Issue template

esanzgar:
Based on the client issue template I suggest add one for the browser-extension.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,52 @@
+---
+name: Bug report
+about: File a bug report for hypothesis' browser-extension.
+title: 'Bug report'
+labels: 'triage_bug'
+assignees: ''
+---
+
+<!--
+Thank you for reporting an issue with the Hypothesis browser-extension! This is our bug
+tracker. If you have any support questions please email support@hypothes.is
+instead.
+
+In order to make it easy for us to help you, please note the following:
+
+1. Only report **bugs** with the Hypothesis browser-extension here. If you want to discuss
+   potential new features, please use the feature request template or contact our developer
+   community here:
+
+       https://groups.google.com/a/list.hypothes.is/forum/#!forum/dev
+
+2. Be sure to check that the bug you're reporting isn't already reported. You
+   can use the GitHub issues search box to help with this.
+
+3. Fill out the information requested by the template below. If you can't supply
+   all the information, don't worry, but do please use the template.
+-->
+
+### Steps to reproduce
+
+1.
+2.
+3.
+
+### Expected behavior
+
+Tell us what should happen.
+
+### Actual behavior
+
+Tell us what happens instead.
+
+### Browser/system information
+
+Tell us which browser, browser version, and operating system you were using when
+you encountered this issue.
+
+### Additional details
+
+If you have other information which might be helpful, please include it. If a
+screenshot will help to illustrate the issue you are reporting, please include
+one.
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Hypothesis Google Forum
+    url: /https://groups.google.com/a/list.hypothes.is/g/dev
+    about: Please ask and answer questions here.
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,39 @@
+---
+name: Feature request
+about: Suggest a new feature for hypothesis' browser-extension.
+title: 'Feature request'
+labels: 'feature_request'
+assignees: ''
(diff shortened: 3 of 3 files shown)


.github/ISSUE_TEMPLATE/custom.md
```diff
@@ -0,0 +1,52 @@
+---
+name: Custom issue template
+about: Describe this issue template's purpose here.
+title: ''
+labels: ''
+assignees: ''
```
robertknight, line 6: If we're going to add metadata, it should have a useful name and description at least. Since this template is for bug reports, I suggest the name and description should reflect that.

Perhaps we could have two separate templates, one for bug reports and one for feature requests or other. If we would prefer that users use other channels for feature requests we can include that information in the feature request template. There are metadata fields for contact links for example. See https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/about-issue-and-pull-request-templates.

robertknight (commented):
GitHub improved their issue template functionality since the client's issue template was set up. At that time a repository could only contain a single issue template and so we had to include a bunch of stuff in an HTML comment which I suspect is often not read.

If we are going to set this up using the [modern issue template functionality](https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/configuring-issue-templates-for-your-repository) then I think we'd want to do something like:

- Have a bug report template (this one) and give it an appropriate filename and YAML metadata at the top
- If we want to re-route certain requests to a different forum or email, use the `contact_links` field in `config.yml`
- For other kinds of issues, either allow users to create blank issues (simplest option) or create a specific template for it

esanzgar:
@robertknight, following your suggestions I created two templates, one for bug reports and one for feature requests. I also added a `config.yml` with contact information.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,52 @@
+---
+name: Bug report
+about: File a bug report for hypothesis' browser-extension.
+title: 'Bug report'
+labels: 'triage_bug'
+assignees: ''
+---
+
+<!--
+Thank you for reporting an issue with the Hypothesis browser-extension! This is our bug
```
robertknight, line 10:
```suggestion
Thank you for reporting an issue with the Hypothesis browser extension! This is our bug
```

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,39 @@
...
+labels: 'feature_request'
+assignees: ''
+---
+
+<!--
+Thank you for requesting a feature for the Hypothesis browser-extension!
+
+In order to make it easy for us to help you, please note the following:
+
+1. Only report **feature_request** with the Hypothesis browser-extension here. If you want to
```
robertknight, line 14: I suggest this bullet point can be removed for the same reasons as in my comment on the bug report template above.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,52 @@
+---
+name: Bug report
+about: File a bug report for hypothesis' browser-extension.
+title: 'Bug report'
+labels: 'triage_bug'
```
robertknight, line 5: I suggest it would make sense to separate the _type_ of issue from whether it needs triaging by a Hypothesis team member or not. So for example I would suggest to use two labels here: `Needs triage` + `bug` (already exists). Also a note on label naming - please make sure the style is consistent with one of the two styles that existing labels use. See https://github.com/hypothesis/browser-extension/labels. They are either "Sentence case" or "all lower case".

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Hypothesis Google Forum
+    url: /https://groups.google.com/a/list.hypothes.is/g/dev
```
robertknight, line 4: I suggest we link to the "Contact Hypothesis" page here instead: https://web.hypothes.is/contact/. This will ensure that it stays up to date with the support team's preferred practices as well as the various communication channels that are available (Slack). Regarding Google Groups, we have two forums, one is a developer-focused forum and the other is [for end users](https://groups.google.com/g/hypothesis-forum). The one for end users is probably more appropriate here and is also monitored by the support team.

robertknight (changes requested):
Thanks for taking another stab at this Eduardo. I left recommendations on various aspects. In particular:

1. I think we should use separate labels from the type of issue vs. whether it has been triaged yet, since we want to indicate "Needs triage" for both bugs and feature requests.
2. We should remove links to the developer forum as it isn't the best place to report bugs any more or find existing documentation for users. Contact links can also get out of date, so I suggest we link to the page on web.hypothes.is that has this information.
3. Of the bullet points in each issue type's text, I think the first is obsoleted by the GitHub UI that makes the user first choose an issue type before they can presented with the template to fill out.

robertknight:
I'm going to close this for now as it has been sitting around for a long time. Feel free to reopen if/when you have time to address the earlier comments.

PR #437, opened 2020-11-11
Improve JSDoc for various classes and types in the background page

robertknight:
Following on from https://github.com/hypothesis/browser-extension/pull/428 this PR fixes issues with and improves the existing JSDoc documentation:

- Fix several JSDoc comments that not start with the correct `/**` text so they are actually parsed as JSDoc
- Convert informal parameter documentation into `@param` so TypeScript can check usage
- Introduce a `@typedef` for the state of each tab so that TS can check usage.
- With the addition of the above typedef, TS can now check that the `state` property (active, inactive, errored) is set to the correct values, so this removes the need for the `states` enum in `tab-state.js`. Since this enum doesn't really have any value any more, I replaced uses of it with equivalent string literals
- Add JSDoc documentation to various methods

In the process TS flagged a couple of places where we assume tabs have an ID and apparently that is not guaranteed (see https://developer.chrome.com/extensions/tabs#type-Tab). I added a check for the one place where we might encounter a tab without an ID.

### DIFF
--- a/src/background/browser-action.js
+++ b/src/background/browser-action.js
@@ -78,11 +78,12 @@ export default function BrowserAction(chromeBrowserAction) {
 
     // display the annotation count on the badge
     if (state.state !== 'errored' && state.annotationCount) {
-      let countLabel;
       let totalString = state.annotationCount.toString();
       if (state.annotationCount > 999) {
         totalString = '999+';
       }
+
+      let countLabel;
       if (state.annotationCount === 1) {
         countLabel = _("There's 1 annotation on this page");
       } else {
@@ -90,6 +91,7 @@ export default function BrowserAction(chromeBrowserAction) {
           'There are ' + totalString + ' annotations on this page'
         );
       }
+
       title = countLabel;
       badgeText = totalString;
     }
--- a/src/background/browser-action.js
+++ b/src/background/browser-action.js
@@ -1,19 +1,16 @@
 import settings from './settings';
-import TabState from './tab-state';
-
-// Cache the tab state constants.
-const states = TabState.states;
 
 // Each button state has two icons one for normal resolution (19) and one
 // for hi-res screens (38).
-const icons = {};
-icons[states.ACTIVE] = {
-  19: 'images/browser-icon-active.png',
-  38: 'images/browser-icon-active@2x.png',
-};
-icons[states.INACTIVE] = {
-  19: 'images/browser-icon-inactive.png',
-  38: 'images/browser-icon-inactive@2x.png',
+const icons = {
+  active: {
+    19: 'images/browser-icon-active.png',
+    38: 'images/browser-icon-active@2x.png',
+  },
+  inactive: {
+    19: 'images/browser-icon-inactive.png',
+    38: 'images/browser-icon-inactive@2x.png',
+  },
 };
 
 // themes to apply to the toolbar icon badge depending on the type of
@@ -34,12 +31,15 @@ function _(str) {
   return str;
 }
 
-/* Controls the display of the browser action button setting the icon, title
+/**
+ * Controls the display of the browser action button setting the icon, title
  * and badges depending on the current state of the tab.
  *
  * BrowserAction is responsible for mapping the logical H state of
  * a tab (whether the extension is active, annotation count) to
  * the badge state.
+ *
+ * @param {chrome.browserAction} chromeBrowserAction
  */
 export default function BrowserAction(chromeBrowserAction) {
   const buildType = settings.buildType;
@@ -51,16 +51,16 @@ export default function BrowserAction(chromeBrowserAction) {
    * @param state - The H state of a tab. See the 'tab-state' module.
    */
   this.update = function (tabId, state) {
-    let activeIcon = icons[states.INACTIVE];
+    let activeIcon = icons.inactive;
     let title = '';
     let badgeText = '';
 
-    if (state.state === states.ACTIVE) {
-      activeIcon = icons[states.ACTIVE];
+    if (state.state === 'active') {
+      activeIcon = icons.active;
       title = 'Hypothesis is active';
-    } else if (state.state === states.INACTIVE) {
+    } else if (state.state === 'inactive') {
       title = 'Hypothesis is inactive';
-    } else if (state.state === states.ERRORED) {
+    } else if (state.state === 'errored') {
       title = 'Hypothesis failed to load';
       badgeText = '!';
     } else {
@@ -68,7 +68,7 @@ export default function BrowserAction(chromeBrowserAction) {
     }
 
     // display the annotation count on the badge
-    if (state.state !== states.ERRORED && state.annotationCount) {
+    if (state.state !== 'errored' && state.annotationCount) {
       let countLabel;
       let totalString = state.annotationCount.toString();
       if (state.annotationCount > 999) {
--- a/src/background/browser-action.js
+++ b/src/background/browser-action.js
@@ -1,5 +1,9 @@
 import settings from './settings';
 
+/**
+ * @typedef {import('./tab-state').State} TabState
+ */
+
 // Each button state has two icons one for normal resolution (19) and one
 // for hi-res screens (38).
 const icons = {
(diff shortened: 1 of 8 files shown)


src/background/hypothesis-chrome-extension.js
```diff
@@ -98,7 +104,7 @@ export default function HypothesisChromeExtension(dependencies) {
     }
 
     chromeTabs.create({ url: settings.serviceUrl + 'welcome' }, function (tab) {
-      state.activateTab(tab.id);
+      state.activateTab(/** @type {number} */ (tab.id));
```
robertknight, line 107: Here it is safe to assume the tab has an ID because we know it is a regular tab.

src/background/hypothesis-chrome-extension.js
```diff
@@ -107,7 +113,9 @@ export default function HypothesisChromeExtension(dependencies) {
     state.load(store.all());
     chromeTabs.query({}, function (tabs) {
       tabs.forEach(function (tab) {
-        onTabStateChange(tab.id, state.getState(tab.id));
+        if (tab.id) {
+          onTabStateChange(tab.id, state.getState(tab.id));
+        }
```
robertknight, line 118: Unlike the previous case I'm not sure that it is safe to assume the tab has an ID, so I added a check. https://developer.chrome.com/extensions/tabs#type-Tab mentions scenarios where it might not.

src/background/sidebar-injector.js
```diff
@@ -48,24 +48,25 @@ function extractContentScriptResult(result) {
...
+ * The SidebarInjector is used to deploy and remove the Hypothesis sidebar
  * from tabs. It also deals with loading PDF documents into the PDF.js viewer
  * when applicable.
  *
- * chromeTabs - An instance of chrome.tabs.
- * dependencies - An object with additional helper methods.
- *   isAllowedFileSchemeAccess: A function that returns true if the user
+ * @param {chrome.tabs} chromeTabs
+ * @param {Object} services
+ * @param {(cb: (allowed: boolean) => void) => void} services.isAllowedFileSchemeAccess -
```
esanzgar, line 58: Shouldn't this property be optional? There is a check in the code to test if exist and is a function.
```suggestion
 * @param {(cb: (allowed: boolean) => void) => void} [services.isAllowedFileSchemeAccess] -
```

src/background/sidebar-injector.js
```diff
@@ -48,24 +48,25 @@ function extractContentScriptResult(result) {
...
+ *   A function that returns true if the user
  *   can access resources over the file:// protocol. See:
```
esanzgar, line 60: Idem as above?

src/background/tab-state.js
```diff
@@ -3,48 +3,61 @@ import { RequestCanceledError } from './errors';
...
- * initialState - An Object of tabId/state keys. Used when loading state
- *   from a persisted store such as localStorage. This will be merged with
- *   the default state for a tab.
- * onchange     - A function that receives onchange(tabId, current).
+ * @param {TabStateMap} initialState - Initial state information for tabs, eg.
+ *   from a persistent store.
+ * @param {(tabId: number, current: undefined|State) => any} onchange -
+ *   Callback invoked when state for a tab changes
  */
 export default function TabState(initialState, onchange) {
```
esanzgar, line 53: Line 77 could we remove if we add this default value.

```suggestion
export default function TabState(initialState, onchange = null) {
```

src/background/tab-state.js
```diff
@@ -3,48 +3,61 @@ import { RequestCanceledError } from './errors';
...
+ * TabState stores the Hypothesis-related state for tabs in the current browser
+ * session.
  *
- * initialState - An Object of tabId/state keys. Used when loading state
- *   from a persisted store such as localStorage. This will be merged with
- *   the default state for a tab.
- * onchange     - A function that receives onchange(tabId, current).
+ * @param {TabStateMap} initialState - Initial state information for tabs, eg.
+ *   from a persistent store.
+ * @param {(tabId: number, current: undefined|State) => any} onchange -
```
esanzgar, line 50: Should it be optional?

src/background/tab-state.js
```diff
@@ -63,14 +76,15 @@ export default function TabState(initialState, onchange) {
...
-   * @param newState - A dictionary mapping tab ID to tab state objects.
+   * @param {TabStateMap} newState - A dictionary mapping tab ID to tab state objects.
    *                   The provided state will be merged with the default
    *                   state for a tab.
    */
   this.load = function (newState) {
+    /** @type {TabStateMap} */
     const newCurrentState = {};
     Object.keys(newState).forEach(function (tabId) {
       newCurrentState[tabId] = Object.assign(
```
esanzgar, line 90: Could we replace `Object.assign` to the spread syntax `{...DEFAULT_STATE, ...newState[tabId]}`

src/background/tab-state.js
```diff
@@ -82,62 +96,108 @@ export default function TabState(initialState, onchange) {
...
+   * Mark a tab as having encountered an error when trying to load Hypothesis into it.
+   *
+   * @param {number} tabId
+   * @param {Error} error
+   */
   this.errorTab = function (tabId, error) {
     this.setState(tabId, {
-      state: states.ERRORED,
+      state: 'errored',
       error: error,
```
esanzgar, line 126: Out of curiosity: `error` object is skip in the serialisation on the `localstorage`?

esanzgar (approved):
I have a couple of minor questions about optional parameters.

src/background/browser-action.js
```diff
@@ -51,24 +51,24 @@ export default function BrowserAction(chromeBrowserAction) {
...
    */
   this.update = function (tabId, state) {
-    let activeIcon = icons[states.INACTIVE];
+    let activeIcon = icons.inactive;
     let title = '';
     let badgeText = '';
 
-    if (state.state === states.ACTIVE) {
-      activeIcon = icons[states.ACTIVE];
+    if (state.state === 'active') {
```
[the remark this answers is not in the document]
robertknight: Yes, I think so. I'll make that change and add document the parameter types as well.

src/background/browser-action.js
```diff
@@ -51,24 +51,24 @@ export default function BrowserAction(chromeBrowserAction) {
    * @param state - The H state of a tab. See the 'tab-state' module.
    */
   this.update = function (tabId, state) {
-    let activeIcon = icons[states.INACTIVE];
+    let activeIcon = icons.inactive;
     let title = '';
```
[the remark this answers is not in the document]
robertknight: I've rewritten this logic now so that each `switch` branch assigns the `activeIcon` and `title` values rather than relying on a fallback.

src/background/browser-action.js
```diff
@@ -51,24 +51,24 @@ export default function BrowserAction(chromeBrowserAction) {
    * @param state - The H state of a tab. See the 'tab-state' module.
```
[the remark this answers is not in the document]
robertknight: Done

src/background/sidebar-injector.js
```diff
@@ -48,24 +48,25 @@ function extractContentScriptResult(result) {
...
+ *   A function that receives a path and returns an absolute
  *   url. See: https://developer.chrome.com/extensions/extension#method-getURL
  */
-export default function SidebarInjector(chromeTabs, dependencies) {
-  dependencies = dependencies || {};
-
-  const isAllowedFileSchemeAccess = dependencies.isAllowedFileSchemeAccess;
-  const extensionURL = dependencies.extensionURL;
-
+export default function SidebarInjector(
```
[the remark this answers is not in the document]
robertknight: Yes - there are several classes in the extension code which still use ES5-style methods of creating a class. I think that's too large of a change for this PR, but it would be a good idea to convert them in future.

src/background/sidebar-injector.js
robertknight: There is a check, but it throws an error if the parameters are missing. These are required parameters though. I'm going to remove the checks since we can now rely on TypeScript to verify that required parameters are passed.

src/background/hypothesis-chrome-extension.js
```diff
@@ -31,24 +32,29 @@ const TAB_STATUS_COMPLETE = 'complete';
  * - https://developer.chrome.com/extensions/tabs
  * - https://developer.chrome.com/extensions/extension
  *
- * dependencies - An object to set up the application.
- *   chromeTabs: An instance of chrome.tabs.
- *   chromeBrowserAction: An instance of chrome.browserAction.
- *   extensionURL: chrome.extension.getURL.
- *   isAllowedFileSchemeAccess: chrome.extension.isAllowedFileSchemeAccess.
+ * @param {Object} services
```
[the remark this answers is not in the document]
robertknight: Thanks for the suggestion. In this context it doesn't appear to make any functional difference - if you try to pass a primitive or null/undefined it will error. I've stuck with `Object` as that's more conventional in our code for this purpose at the moment, a convention that I think came from the [TS handbook](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns).

I agree we should prefer `object` in cases where it does make a difference though, assuming we can't be more specific about the object's properties.

src/background/tab-state.js
robertknight: Indeed. I've gone one step further and just marked the parameter as optional. There was no particular reason to initialize the `onchange` field of this class to `null` rather than `undefined` if not set from what I can see.

src/background/tab-state.js
robertknight: Yes, I'll update this.

src/background/tab-state.js
robertknight: Yes. I'm going to leave that out of scope for this PR, but use of `Object.assign` is one of many pre-modern ES code patterns that can be updated whenever someone has the time.

src/background/tab-state.js
robertknight: Yes. The `TabStore.set` method only persists the `state` and `annotationCount` properties. It is possible this could cause problems. I haven't checked.

PR #439, opened 2020-11-17
Prevent error if tab.id is undefined

esanzgar:
The following error is quite frequent in the console:

```
Error handling response: TypeError: Cannot read property 'id' of undefined at updateTabDocument
```

This is caused by async calls to `onTabStateChange`. By the time these callback are executed the tab can be already deleted.

An example of this happens when login in the client. The pop up window is closed and later `onTabStateChange` called.

### DIFF
--- a/src/background/hypothesis-chrome-extension.js
+++ b/src/background/hypothesis-chrome-extension.js
@@ -120,14 +120,28 @@ export default function HypothesisChromeExtension({
     });
   }
 
+  /**
+   *
+   * @param {number} tabId
+   * @param {import('./tab-state').State|undefined} current
+   */
   function onTabStateChange(tabId, current) {
     if (current) {
-      browserAction.update(tabId, current);
-      chromeTabs.get(tabId, updateTabDocument);
+      chromeTabs.get(tabId, tab => {
+        // This error is raised if the tab doesn't exist.
+        if (chrome.runtime.lastError) {
+          state.clearTab(tabId);
+          return;
+        }
 
-      if (!state.isTabErrored(tabId)) {
-        store.set(tabId, current);
-      }
+        browserAction.update(tabId, current);
+
+        updateTabDocument(tab);
+
+        if (!state.isTabErrored(tabId)) {
+          store.set(tabId, current);
+        }
+      });
     } else {
       store.unset(tabId);
     }
@@ -219,23 +233,29 @@ export default function HypothesisChromeExtension({
     state.clearTab(tabId);
   }
 
-  // installs or uninstalls the sidebar from a tab when the H
-  // state for a tab changes
+  /**
+   * Installs or uninstalls the sidebar from a tab when the H
+   * state for a tab changes
+   *
+   * @param {chrome.tabs.Tab} tab
+   */
   function updateTabDocument(tab) {
+    const tabId = /** @type {number} */ (tab.id);
+
     // If the tab has not yet finished loading then just quietly return.
-    if (!state.getState(tab.id).ready) {
-      return Promise.resolve();
+    if (!state.getState(tabId).ready) {
+      return;
     }
 
-    const isInstalled = state.getState(tab.id).extensionSidebarInstalled;
-    if (state.isTabActive(tab.id) && !isInstalled) {
+    const isInstalled = state.getState(tabId).extensionSidebarInstalled;
+    if (state.isTabActive(tabId) && !isInstalled) {
       // optimistically set the state flag indicating that the sidebar
       // has been installed
-      state.setState(tab.id, {
+      state.setState(tabId, {
         extensionSidebarInstalled: true,
       });
 
-      const { directLinkQuery } = state.getState(tab.id);
+      const { directLinkQuery } = state.getState(tabId);
 
       const config = {
         // Configure client to load assets and sidebar app from extension.
@@ -253,15 +273,15 @@ export default function HypothesisChromeExtension({
       // that modify the URL fragment as they load. See commit 3143ca27e05d.
       Object.assign(config, directLinkQuery);
 
-      return sidebar
+      sidebar
         .injectIntoTab(tab, config)
         .then(function () {
           // Clear the direct link once H has been successfully injected
-          state.setState(tab.id, { directLinkQuery: undefined });
+          state.setState(tabId, { directLinkQuery: undefined });
         })
         .catch(function (err) {
           if (err instanceof errors.AlreadyInjectedError) {
-            state.setState(tab.id, {
+            state.setState(tabId, {
               state: 'inactive',
               extensionSidebarInstalled: false,
             });
@@ -272,16 +292,14 @@ export default function HypothesisChromeExtension({
               url: tab.url,
             });
           }
-          state.errorTab(tab.id, err);
+          state.errorTab(tabId, err);
         });
-    } else if (state.isTabInactive(tab.id) && isInstalled) {
-      return sidebar.removeFromTab(tab).then(function () {
-        state.setState(tab.id, {
+    } else if (state.isTabInactive(tabId) && isInstalled) {
+      sidebar.removeFromTab(tab).then(function () {
+        state.setState(tabId, {
           extensionSidebarInstalled: false,
         });
       });
-    } else {
-      return Promise.resolve();
     }
   }
 
@@ -292,12 +310,14 @@ export default function HypothesisChromeExtension({
       return;
     }
 
+    // If user disabled the badge count, this call to `sync.get` will
+    // return `{ badge: false}`
     chromeStorage.sync.get(
       {
-        badge: true,
+        badge: true, // the default value `true` is returned only if `badge` is not yet set.
       },
-      function (items) {
-        if (items.badge) {
+      ({ badge }) => {
+        if (badge) {
           state.updateAnnotationCount(tabId, url);
         }
       }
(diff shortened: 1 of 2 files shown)


src/background/hypothesis-chrome-extension.js
```diff
@@ -219,11 +219,15 @@ export default function HypothesisChromeExtension({
...
+  /**
+   * installs or uninstalls the sidebar from a tab when the H
+   * state for a tab changes
+   *
+   * @param {chrome.tabs.Tab} tab
+   */
   function updateTabDocument(tab) {
     // If the tab has not yet finished loading then just quietly return.
-    if (!state.getState(tab.id).ready) {
+    if (tab?.id === undefined || !state.getState(tab.id).ready) {
```
robertknight, line 230: When does this situation happen where `tab` is undefined? Is the calling code doing something wrong?

src/background/hypothesis-chrome-extension.js
```diff
@@ -219,11 +219,15 @@ export default function HypothesisChromeExtension({
     state.clearTab(tabId);
   }
 
-  // installs or uninstalls the sidebar from a tab when the H
-  // state for a tab changes
+  /**
+   * installs or uninstalls the sidebar from a tab when the H
+   * state for a tab changes
+   *
+   * @param {chrome.tabs.Tab} tab
```
robertknight, line 226: The JSDoc params are inconsistent with the code here. `tab` is marked as not-null in the JSDoc type but then you're checking for the case where it is nullish below. Either we should avoid calling this function if we don't have a tab, or we should make it clear that the argument might be nullish and the function needs to handle that.

src/background/hypothesis-chrome-extension.js
esanzgar: When you query for a tab that doesn't exist:

```
chrome.tabs.get(1, a => console.log(a))
undefined
Unchecked runtime.lastError: No tab with id: 1.
```

The callback get executed and an error is raised.

src/background/hypothesis-chrome-extension.js
esanzgar: Yes, I agree about the inconsistency.

src/background/hypothesis-chrome-extension.js
esanzgar: This is related to #320

src/background/hypothesis-chrome-extension.js
```diff
@@ -123,6 +123,9 @@ export default function HypothesisChromeExtension({
   function onTabStateChange(tabId, current) {
     if (current) {
       browserAction.update(tabId, current);
+
+      // raises an `Unchecked runtime.lastError: No tab with id: X.`,
+      // where X is a tab that doesn't exist
```
esanzgar, line 128: I wonder if we should catch the `Error` and clear the tab.

src/background/hypothesis-chrome-extension.js
```diff
@@ -120,14 +120,28 @@ export default function HypothesisChromeExtension({
...
   }
 
+  /**
+   *
+   * @param {number} tabId
+   * @param {import('./tab-state').State|undefined} current
+   */
   function onTabStateChange(tabId, current) {
     if (current) {
-      browserAction.update(tabId, current);
```
esanzgar, line 125: This is moved inside the callback below (`chromeTabs.get`) and only executed if there is no errors getting the tab.

src/background/hypothesis-chrome-extension.js
```diff
@@ -253,15 +273,15 @@ export default function HypothesisChromeExtension({
       // that modify the URL fragment as they load. See commit 3143ca27e05d.
       Object.assign(config, directLinkQuery);
 
-      return sidebar
```
esanzgar, line 256: Nobody is consuming this promise.

src/background/hypothesis-chrome-extension.js
```diff
@@ -272,16 +292,14 @@ export default function HypothesisChromeExtension({
...
-    } else if (state.isTabInactive(tab.id) && isInstalled) {
-      return sidebar.removeFromTab(tab).then(function () {
-        state.setState(tab.id, {
+    } else if (state.isTabInactive(tabId) && isInstalled) {
+      sidebar.removeFromTab(tab).then(function () {
+        state.setState(tabId, {
           extensionSidebarInstalled: false,
         });
       });
-    } else {
```
esanzgar, line 283: It is not needed

src/background/hypothesis-chrome-extension.js
```diff
@@ -219,23 +233,29 @@ export default function HypothesisChromeExtension({
     state.clearTab(tabId);
   }
 
-  // installs or uninstalls the sidebar from a tab when the H
-  // state for a tab changes
+  /**
+   * installs or uninstalls the sidebar from a tab when the H
```
robertknight, line 237:
```suggestion
   * Installs or uninstalls the sidebar from a tab when the H
```

src/background/hypothesis-chrome-extension.js
robertknight: This is true. I guess the original reason it was done this way was to make it possible for any callers to tell when `updateTabDocument` had finished whatever work it is going to do.

src/background/hypothesis-chrome-extension.js
```diff
@@ -292,12 +310,14 @@ export default function HypothesisChromeExtension({
       return;
     }
 
+    // If user disabled the badge count, this call to `sync.get` will
+    // return `{ badge: false}`
```
robertknight, line 314: Is the result after an initial install just `{}`?

tests/background/hypothesis-chrome-extension-test.js
```diff
@@ -55,6 +55,7 @@ describe('HypothesisChromeExtension', function () {
   }
 
   beforeEach(function () {
+    global.chrome = { runtime: { lastError: false } };
```
robertknight, line 58: This state could potentially leak into other tests. I suggest we make sure this change is undone in an `afterEach` callback.

src/background/hypothesis-chrome-extension.js
```diff
@@ -120,14 +120,28 @@ export default function HypothesisChromeExtension({
...
+   *
+   * @param {number} tabId
+   * @param {import('./tab-state').State|undefined} current
+   */
   function onTabStateChange(tabId, current) {
     if (current) {
-      browserAction.update(tabId, current);
-      chromeTabs.get(tabId, updateTabDocument);
+      chromeTabs.get(tabId, tab => {
+        // This error is raised if the tab doesn't exists.
```
robertknight, line 131:
```suggestion
        // This error is raised if the tab doesn't exist.
```

Are there any other reasons this error might be raised that you know of?

src/background/hypothesis-chrome-extension.js
```diff
@@ -120,14 +120,28 @@ export default function HypothesisChromeExtension({
...
+  /**
+   *
+   * @param {number} tabId
+   * @param {import('./tab-state').State|undefined} current
+   */
   function onTabStateChange(tabId, current) {
     if (current) {
-      browserAction.update(tabId, current);
-      chromeTabs.get(tabId, updateTabDocument);
+      chromeTabs.get(tabId, tab => {
```
robertknight, line 130: I wondered here whether higher-level / earler code had made an error if we call `chromeTabs.get` for a tab that doesn't exist. From what I can see the sequence of events that happens when logging in via the Hypothesis client's authorization popup is:

1. User submits the form
2. `onTabUpdated` is called for the popup window with `{status: "complete"}` in the `changeInfo`. This updates the tab's state.
3. The state update triggers a call to `onTabStateChange` which in turn calls `chrome.tabs.get(...)`
4. `onTabRemoved` is called for the popup window, which removes the tab's state from the local store
5. The async callback for the `chrome.tabs.get(...)` call in step 3 fires.

So in this case the tab ID was valid at the point where `chrome.tabs.get` was called, but not when the callback got processed.

I think it would be worth noting in the comment you added below that `chrome.tabs.get` can fail even if the tab ID was valid (or at least, we hadn't received a notification yet that the tab had been removed) when `chrome.tabs.get` was called.

robertknight (approved):
The change here makes sense. It took me a while to check the sequence of events that was happening to make sure that the code calling `onTabStateChange` hadn't done something wrong. Testing the case of the client's auth popup I can see though that the tab ID is valid, at least as far as the extension is aware, at the time when `onTabStateChange` is called, but either the extension has not been notified that the tab has been removed or the tab ID becomes invalid sometime between then and the `chrome.tabs.get` callback being handled.

It seems to me that we should be handling this possibility whenever `chrome.tabs.get` is called in the extension, and in fact any API that tries to access or modify a tab. I found at least one other place in the extension where we are calling `chrome.tabs.get` and not checking `chrome.runtime.lastError`.

On a related note, we have a helper in `src/background/util.js` that can convert callback-based Chrome extension APIs into ones that return promises so that we can write easier-to-follow code for working with them. This might also help us to migrate to promised-based APIs provided by Chrome in future ([related Chrome issue](https://bugs.chromium.org/p/chromium/issues/detail?id=328932)).

src/background/hypothesis-chrome-extension.js
esanzgar: Yes, `{}`.

tests/background/hypothesis-chrome-extension-test.js
esanzgar: Will do

src/background/hypothesis-chrome-extension.js
esanzgar: I don't think the ID is valid when `chrome.tabs.get is called`. 

There are a few more callbacks fired:
6. The async callback after the badge count is fetched (which is now delayed by at least 500ms).
7. The async callback after calling `onTabRemoved` on step 4 (which triggers a state update).

This recursive callbacks are not ideal. I would suggest to change this to be more lineal and easy to follow.

PR #450, opened 2020-12-02
Load to memory only information about the current tabs

esanzgar:
The storage is accumulating information from very old tabs. This is becuase when the extension is restarted it loads all the information in the storage.

This change will enable to load only the information about the current tabs.

### DIFF
--- a/src/background/hypothesis-chrome-extension.js
+++ b/src/background/hypothesis-chrome-extension.js
@@ -109,21 +109,21 @@ export default function HypothesisChromeExtension({
   };
 
   function restoreSavedTabState() {
-    store.reload();
-    state.load(store.all());
     chromeTabs.query({}, function (tabs) {
-      tabs.forEach(function (tab) {
-        if (tab.id) {
-          onTabStateChange(tab.id, state.getState(tab.id));
-        }
+      const tabIds = tabs
+        .filter(tab => tab.id !== undefined)
+        .map(({ id }) => /** @type {number} */ (id));
+      store.reload(tabIds);
+      state.load(store.all());
+      tabIds.forEach(tabId => {
+        onTabStateChange(tabId, state.getState(tabId));
       });
     });
   }
 
   /**
-   *
    * @param {number} tabId
-   * @param {import('./tab-state').State|undefined} current
+   * @param {import('./tab-state').State | undefined} current
    */
   function onTabStateChange(tabId, current) {
     if (current) {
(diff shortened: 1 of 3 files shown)


src/background/hypothesis-chrome-extension.js
```diff
@@ -109,17 +109,20 @@ export default function HypothesisChromeExtension({
...
 
   function restoreSavedTabState() {
-    store.reload();
-    state.load(store.all());
     chromeTabs.query({}, function (tabs) {
-      tabs.forEach(function (tab) {
-        if (tab.id) {
-          onTabStateChange(tab.id, state.getState(tab.id));
-        }
+      const tabIds = tabs.map(({ id }) => /** @type {number} */ (id));
```
robertknight, line 113: Is this cast safe? Can `chrome.tabs.query` return tabs that do not have an ID?

src/background/hypothesis-chrome-extension.js
```diff
@@ -109,17 +109,20 @@ export default function HypothesisChromeExtension({
...
+      tabIds.forEach(tabId => {
+        onTabStateChange(tabId, state.getState(tabId));
       });
     });
   }
 
+  /**
+   * @param {number} tabId
+   * @param {import('./tab-state').State | undefined} current
+   */
```
robertknight, line 125: 👍 for filling out missing types while updating existing code.

robertknight (commented):
I had a query about whether we can rely on tabs returned by `chrome.tabs.query` always having an ID.

src/background/hypothesis-chrome-extension.js
esanzgar: I believe it is safe, but to make 100% I will change it in this way:
```suggestion
      const tabIds = tabs.filter(tab => tab.id).map(({ id }) => id));
```

tests/background/tab-store-test.js
```diff
@@ -14,16 +14,34 @@ describe('TabStore', function () {
...
-      store.reload();
+      store.reload([1, 10]);
+
+      assert.deepEqual(store.all(), {
+        1: { state: 'active', annotationCount: 3 },
+      });
+    });
+
+    it('returns empty object if an error is encountered while loading', () => {
+      fakeLocalStorage.data.state = 'not a valid JSON';
```
robertknight, line 38:
```suggestion
      fakeLocalStorage.data.state = 'not valid JSON';
```

robertknight (approved):
LGTM 👍

PR #451, opened 2020-12-02
Remove support for old format in the localStorage

esanzgar:
5 years ago a more flexible, multi-key object was introduced to capture the relevant information about the tabs. The old format, a `"inactive"|"active"` string was retired.

On every extension update the code that runs the conversion is executed. By now, every localStorage is in the new format, therefore this legacy code can safely be retired.

### DIFF
--- a/src/background/tab-store.js
+++ b/src/background/tab-store.js
@@ -40,16 +40,7 @@ export default function TabStore(storage) {
     try {
       local = {};
       const loaded = JSON.parse(storage.getItem(key));
-      Object.keys(loaded).forEach(function (key) {
-        // ignore tab state saved by earlier versions of
-        // the extension which saved the state as a {key: <state string>}
-        // dict rather than {key: <state object>}
-        if (typeof loaded[key] === 'string') {
-          local[key] = { state: loaded[key] };
-        } else {
-          local[key] = loaded[key];
-        }
-      });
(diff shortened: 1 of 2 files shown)


src/background/tab-store.js
```diff
@@ -40,16 +40,7 @@ export default function TabStore(storage) {
...
-        // ignore tab state saved by earlier versions of
-        // the extension which saved the state as a {key: <state string>}
-        // dict rather than {key: <state object>}
-        if (typeof loaded[key] === 'string') {
-          local[key] = { state: loaded[key] };
-        } else {
-          local[key] = loaded[key];
-        }
-      });
+      Object.keys(loaded).forEach(key => (local[key] = loaded[key]));
```
robertknight, line 43: It would be good to add a test to check what happens if the parsed JSON data is not an object. It looks like such code would always result in the fallback `local = local || {}` below initializing it to an empty object.

robertknight (approved):
LGTM. I suggested adding a test for what happens if the parsed JSON data is an unexpected non-object value.

src/background/tab-store.js
esanzgar: I added a similar test in this other [PR](https://github.com/hypothesis/browser-extension/blob/d9cdf54ae6d669b842e915232d6c2ed2db6e43b3/tests/background/tab-store-test.js#L39)

Is this what you suggested?

PR #452, opened 2020-12-02
Fix Sentry problem in production and qa

esanzgar:
I believe a recent change in Babel has enabled the import of modules without `default` export.

With those changes if the import was like this:

`import * Raven from 'raven-js`

then functions needed to be invoked like this:

`Raven.default.config` (notice the `default` property)

I enabled `"allowSyntheticDefaultImports": true` in tsconfig.json to match the Babel behaviour.

### DIFF
--- a/src/tsconfig.json
+++ b/src/tsconfig.json
@@ -7,12 +7,9 @@
     "noEmit": true,
     "strict": true,
     "noImplicitAny": false,
-    "target": "ES2020"
+    "target": "ES2020",
+    "allowSyntheticDefaultImports": true
   },
-  "include": [
-    "**/*.js"
-  ],
-  "exclude": [
-    "vendor/"
-  ]
+  "include": ["**/*.js"],
+  "exclude": ["vendor/"]
 }
--- a/src/background/raven.js
+++ b/src/background/raven.js
@@ -6,7 +6,7 @@
  * version to be provided via the app's settings object.
  */
 
-import * as Raven from 'raven-js';
+import Raven from 'raven-js';
 
 /**
  * Returns the input URL if it is an HTTP URL or the filename part of the URL


esanzgar:
In order to test this in development, the property `sentryPublicDSN` should be added in the settings file that is been used.

src/tsconfig.json
```diff
@@ -7,12 +7,9 @@
     "noEmit": true,
     "strict": true,
     "noImplicitAny": false,
-    "target": "ES2020"
+    "target": "ES2020",
+    "allowSyntheticDefaultImports": true
```
robertknight, line 11: Are there downsides to enabling this globally? If we incorrectly try to use a default import for a modern package which doesn't have a default export, will that cause problems?

robertknight (commented):
I had a query about whether enabling this change globally could cause any coding mistakes to go unnoticed in future.

src/tsconfig.json
esanzgar: In my opinion there is no downside, but the opposite it would prevent future issues. This change in the tsconfig.json is matching what Babel is doing.

robertknight: OK, let's go with this for now. The code that Babel generates to enable handle imports from CommonJS modules adds bloat to the generated bundle so I'd like to eventually turn both this type checking option and the Babel compatibility code off.

robertknight (approved):
I'm OK with enabling this option for the moment to bring TypeScript and Babel into sync, but I think we should avoid enabling it for other projects where it has not been needed. In fact if we can tell Babel not to generate the compatibility code for CommonJS imports and thus reduce bundle size a bit, that would be better.

esanzgar:
Thanks, Robert! I will disable the `allowSyntheticDefaultImports` when we adopt `@sentry/browser`

PR #457, opened 2020-12-07
Start using  @sentry/browser instead of raven-js

esanzgar:
`raven-js` has been replaced by `@sentry/browser`

Previous filtering options have been disabled and will be re-enabled
progressively.

`allowSyntheticImports` has been disabled as per request of
@robertknight (see: https://github.com/hypothesis/browser-extension/pull/452#pullrequestreview-545048998)


robertknight:
Hi Eduardo - I'm going to suggest we close this PR for the moment and come back to it later. It's quite a bit of a context switch at the moment and we don't have an urgent need to ship this. Note that you can re-open closed PRs at any time in GitHub.

PR #460, opened 2020-12-10
WIP: convert constructor functions to ES2015 classes

esanzgar:


esanzgar:
This was only for demonstration purpose of what it takes to convert from constructor functions to ES2015 classes.

PR #469, opened 2020-12-17
Add HTML file for notebook app and set `notebookAppUrl` in client config

robertknight:
This PR fixes an issue where a locally-built production extension fails to load, with an error occurring due to a missing `notebookAppUrl` config key in the client boot script. It adds rules to the Makefile to generate a `notebook.html` file and configures the client to set it as the URL of the notebook app.

While this PR fixes an immediate problem, there is a higher-level choice of whether we want the extension to "include" its own Notebook app and show that inside an iframe. We could instead make the "Open notebook" link just open https://hypothes.is/notebook (or equivalent URL depending on the h backend the extension is using) in a new tab. I figure we can merge this now and revisit later if we decide to go down that route.

---

To test locally:

1. Remove any existing install of the production Hypothesis client from the Chrome Web Store
2. Build production extension locally with `make SETTINGS_FILE=settings/chrome-prod.json`
3. Load the generated unpackaged extension from chrome://extensions
4. Go to a web page and try to activate the extension

On master the extension fails to activate due to missing `notebookAppUrl` config. On this branch the extension loads and the notebook can be launched via the "Open Notebook" menu item.

When the notebook app does launch, it currently just loads the sidebar app in the iframe, instead of the extension. This is because the client's router is configured to recognize `/notebook` as the URL to set the initial route and the actual URL in the  extension is `/notebook.html`. It is TBD how we'll resolve this. The easiest option would just be treat both `/notebook` and `/notebook.html` as paths that trigger the notebook app route in the client.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ extension: build/extension.bundle.js
 extension: build/manifest.json
 extension: build/client/build
 extension: build/client/app.html
+extension: build/client/notebook.html
 extension: build/settings-data.js
 extension: build/unload-client.js
 extension: build/pdfjs-init.js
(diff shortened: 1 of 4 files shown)


lyzadanger (approved):
I used this branch locally a few weeks ago and it worked just fine. Thank you for dealing with this! And I agree that some strategic thinking will be necessary to complement this tactical fix.

PR #478, opened 2021-01-08
Add code coverage

esanzgar:
On the first commit, I added the configuration and libraries to support code coverage via istanbul.

On the second commit, I fixed one issue one of the tests. As explained in the commit:

Istanbul is inserting an object of the form `cov_xxxx` in each function
to keep track of the lines that are been executed. However, because on
one test the code is run through `eval` in another context, it complains
about the `cov_xxxx` reference not been found.

Pragmatically, I circumvented this issue by adding the minimal number of
`/* istanbul ignore next */` comments on the code that run through the
eval function.

Alternatively, I have tried unsuccessfully to replace `eval` by
`Function` on the same window, so the context is not lost, as suggested
here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#Never_use_eval!


robertknight:
I rebased this PR on master to resolve a `package.json` conflict. I found that I was able to remove two of the `/* istanbul ignore next */` comments, on functions which are executed in the background page rather than in the context of the page being annotated.

robertknight (approved):
LGTM

PR #481, opened 2021-01-12
Add 'make dev' target to continuously build the extension

esanzgar:
To facilitate development, `make dev` target has been added. This command
watches the dependency files (via gulp) and builds the browser-extension
when those files are modified.

This new commands support custom setting files: `make dev [SETTING_FILE=custom.js`.

In addition, `make sure` and `make help` targets have been added to make
this project behave similarly to the client project.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,10 +7,31 @@ MUSTACHE := node_modules/.bin/mustache
 PRETTIER := node_modules/.bin/prettier
 
 .PHONY: default
-default: extension
+default: help
+
+.PHONY: help
+help:
+	@echo "make help              Show this help message"
+	@echo "make dev               Watch for changes and build the browser-extension"
+	@echo "make build             Create a build of the browser-extension"
+	@echo "make lint              Run the code linter(s) and print any warnings"
+	@echo "make checkformatting   Check code formatting"
+	@echo "make format            Automatically format code"
+	@echo "make test              Run the unit tests once"
+	@echo "make sure              Make sure that the formatter, linter, tests, etc all pass"
+	@echo "make clean             Delete development artefacts (cached files, "
+	@echo "                       dependencies, etc)"
+
+.PHONY: build
+build: node_modules/.uptodate extension
+
+.PHONY: dev
+dev: node_modules/.uptodate
+	yarn gulp watch
 
 .PHONY: clean
 clean:
+	rm -f node_modules/.uptodate
 	rm -rf build/* build/.settings.json build/.*.deps
 	rm -rf dist/*
 
@@ -92,5 +113,12 @@ format:
 .PHONY: test
 test:
 	yarn test
+
+.PHONY: sure
+sure: checkformatting lint test
+
+node_modules/.uptodate: package.json yarn.lock
+	yarn install
+	@touch $@
 
 -include build/.*.deps
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,0 +1,29 @@
+/* eslint-disable no-console */
+/* eslint-env node */
+
+'use strict';
+
+const { spawn } = require('child_process');
+
+const { parallel, watch } = require('gulp');
+
+function build(cb) {
+  const make = spawn('make', ['build'], { stdio: 'inherit' });
+  make.on('close', code => {
+    if (code !== 0) {
+      cb(new Error(`make exited with status ${code}`));
+    } else {
+      cb(null);
+    }
+  });
+}
+
+function watchClient() {
+  watch('node_modules/hypothesis', { events: 'all' }, build);
+}
+
+function watchSrc() {
+  watch('src', { events: 'all' }, build);
+}
+
+exports.watch = parallel(build, watchClient, watchSrc);
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -91,16 +91,16 @@ node {
 
     stage('Test') {
         nodeEnv.inside("-e HOME=${workspace}") {
-          sh "make checkformatting lint test"
+          sh "make sure"
         }
     }
 
     stage('Build Packages') {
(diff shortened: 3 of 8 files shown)


gulpfile.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable no-console */
+/* eslint-env node */
+
+'use strict';
+
+const { parallel, watch } = require('gulp');
+const { exec } = require('child_process');
```
robertknight, line 7:
```suggestion
const { exec } = require('child_process');

const { parallel, watch } = require('gulp');
```

Our general convention is to organize imports in this format:

```js
// Node imports (or other libraries that are native to the browser/Node)

// Third party imports

// Local imports
```

gulpfile.js
```diff
@@ -0,0 +1,25 @@
...
+function build(cb) {
+  exec('make build', (error, stdout, stderr) => {
+    console.error(stderr);
+    console.log(stdout);
+    cb(error);
+  });
+}
+
+function watchClient() {
+  watch('node_modules/hypothesis', { events: 'all' }, build);
```
robertknight, line 18: Do we need `events: 'all'` here? According to https://gulpjs.com/docs/en/getting-started/watching-files/#watched-events the default behavior is to:

>  executes tasks whenever a file is created, changed, or deleted. 

That's probably adequate.

robertknight (changes requested):
The basic approach here looks good. Although there are probably simpler alternatives to Gulp for watching files and running commands in response, it seems like a reasonable choice here since we're all familiar with it. The main issue I flagged up is that the Jenkinsfile contains references to the bare `make` command which needs to be updated to `make build`.

Makefile
```diff
@@ -7,7 +7,27 @@ MUSTACHE := node_modules/.bin/mustache
 PRETTIER := node_modules/.bin/prettier
 
 .PHONY: default
-default: extension
+default: help
```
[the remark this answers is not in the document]
esanzgar: I fixed these references in the Jenkinsfile.

Makefile
```diff
@@ -7,7 +7,27 @@ MUSTACHE := node_modules/.bin/mustache
...
+	@echo "make test              Run the unit tests once"
+	@echo "make sure              Make sure that the formatter, linter, tests, etc all pass"
+	@echo "make clean             Delete development artefacts (cached files, "
+	@echo "                       dependencies, etc)"
+
+.PHONY: build
+build: extension
+
+.PHONY: dev
+dev:
```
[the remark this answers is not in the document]
esanzgar: Good idea, thanks

gulpfile.js
esanzgar: You are probably right. Because `./node_module/hypothesis` is sometimes a soft link (because of `yarn link`) I wanted to catch the `unlink` event which is available in `events: all`:

> The available events are 'add', 'addDir', 'change', 'unlink', 'unlinkDir', 'ready', 'error'. Additionally 'all' is available, which represents all events other than 'ready' and 'error'.

esanzgar:
I addressed all your comments, @robertknight.

This is ready for a second pass.

gulpfile.js
esanzgar: I confirm that `events: all` it is necessary for example to detect the addition of new files to it triggers the transpilation.

gulpfile.js
```diff
@@ -0,0 +1,26 @@
...
+const { exec } = require('child_process');
+
+const { parallel, watch } = require('gulp');
+
+function build(cb) {
+  exec('make build', (error, stdout, stderr) => {
+    console.error(stderr);
+    console.log(stdout);
+    cb(error);
+  });
```
robertknight, line 15: `make build` can take a few seconds to run and currently the output is only shown at the end. A possibility to improve this is to use `spawn` with `{ stdio: 'inherit' }` so that output from the command is forwarded as it is emitted by make:

```js
function build(cb) {
  const make = spawn('make', ['build'], { stdio: 'inherit' });
  make.on('close', code => {
    if (code !== 0) {
      cb(new Error(`make exited with status ${code}`));
    } else {
      cb(null);
    }
  });
}
```

gulpfile.js
```diff
@@ -0,0 +1,26 @@
...
+}
```
robertknight, line 16: A related issue here is that in development it is common to run `make build` with different values for `SETTINGS_FILE`. It would be useful to have a documented way to run `make dev` with different values for this argument. One option I can suggest is that `make dev SETTINGS_FILE=<value>` would forward the argument to gulp that would then forward it to `make build`.

gulpfile.js
```diff
@@ -0,0 +1,26 @@
...
+
+function watchClient() {
+  watch('node_modules/hypothesis', { events: 'all' }, build);
+}
```
robertknight, line 20: My typical workflow for updating the Hypothesis client involves using [yalc](https://github.com/whitecolor/yalc) to "publish" the client on my local system and then use it within the browser extension repo. Running `yalc link hypothesis` doesn't trigger a rebuild currently.

This isn't a workflow that we recommend in the project README so I think that issue is OK for now, but I'll probably take a look when I have time.

robertknight (approved):
I added a suggestion regarding forwarding output of `make build` to the console. I think this is good to go though. Can you please do a QA + prod release after this lands make sure the end-to-end deployment workflow is working with the changes in the Jenkinsfile.

gulpfile.js
esanzgar: Great suggestion!

gulpfile.js
esanzgar: I have not tried with `yalc`, I currently use `yarn link`. I can take a look at this on a follow up.

PR #531, opened 2021-03-22
Configure dev settings

oltyx:
Just copied the production settings in it.


robertknight:
I assume this PR was submitted by accident? The settings in settings/chrome-dev.json are intentionally different from the production ones.

PR #612, opened 2021-07-22
Replace query-string by `URL.searchParams`

robertknight:
Remove a dependency by using `URL.searchParams` to extract a query param from a URL.

This also resolves a theoretical issue where the `originalURL` variable could end up being an array rather than a string if the `file` query param was repeated in the URL, although I don't know of non-contrived scenario (eg. user manually editing URL bar) in which that might happen.


lyzadanger (approved):
This is a nice change; I like the red in this diff!

PR #624, opened 2021-08-06
Simplify `content_security_policy` in manifest

robertknight:
The Hypothesis client no longer uses Google Analytics, inline JavaScript
(eval) or inline fonts.

The default value of `content_security_policy` would now work, although
we can actually be more strict. The default policy allows additional
script sources (blob, filesystem) which we don't need.

### DIFF
--- a/src/manifest.json.mustache
+++ b/src/manifest.json.mustache
@@ -54,7 +54,8 @@
     "storage",
     "tabs"
   ],
-  "content_security_policy": "script-src 'self' 'unsafe-eval' https://www.google-analytics.com; object-src 'self'; font-src 'self' data:;",
+
+  "content_security_policy": "script-src 'self'; object-src 'self'",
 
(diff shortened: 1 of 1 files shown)


esanzgar (approved):
Looks OK.

We don't need fonts rule?

Maybe at some point we should migrate to version 3 of the manifest.

robertknight:
> We don't need font-src rule? The client contains a number of fonts.

We need to allow fonts served from the extension itself, but we don't need `data:` as a font source as far as I'm aware. We don't have any font-src rules in the CSP policy used by the embedded client (https://hypothes.is/app.html). I've tested that math equations still render.

robertknight:
> Maybe at some point we should migrate to version 3 of the manifest.

Yes we should. I don't know exactly when we'll fit that work in, but it is definitely something we intend to do.

PR #630, opened 2021-08-10
Fix PDF.js viewer failing to load without unsafe-eval

robertknight:
Fix loading of the PDF.js viewer (viewer.html) after removing the
unsafe-eval permission from the extension's content security policy.

It turns out that an async/await polyfill (regenerator) in PDF.js's
source bundles relied on this in a fallback code path if assigning a
global `regeneratorRuntime` variable failed:

```js
// Simplified version of code in various PDF.js code bundles
try {
  // Code that fails if script is run in "strict" mode and `regeneratorRuntime` is not already defined
  regeneratorRuntime = <value>
} catch (...) {
  // Fallback code path that relies on `eval`.
  new Function('r', 'regeneratorRuntime = r')(<value>)
}
```

Fix the issue by predefining the `regeneratorRuntime` global before
PDF.js loads. We should be able to remove this in future after upgrading
to a release of PDF.js that uses native async/await. The change in this PR
is simpler than a major PDF.js version update though.

 - Add `pdfjs-setup-env.js` as a place to prepare the global environment
   before PDF.js loads and predefine `regeneratorRuntime` in it

 - Load pdfjs-setup-env.js before PDF.js in viewer.html

 - Configure PDF.js to use a custom wrapper around pdf.worker.js in
   pdfjs-init.js. This wrapper, in pdfjs-worker-init.js, loads
   pdfjs-setup-envjs before the real worker script.

----

**Testing:**

- Build the extension locally, check that you can view and annotate a PDF with it
- When viewing a PDF, check that there are no worker-related warnings or errors in the console (eg. messages about using a "fake worker")

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,11 +88,7 @@ build/settings-data.js: src/settings-data.js.mustache build/client build/.settin
 	tools/template-context-settings.js build/.settings.json | $(MUSTACHE) - $< >$@
 build/unload-client.js: src/unload-client.js
 	cp $< $@
-build/pdfjs-init.js: src/pdfjs-init.js
-	cp $< $@
-build/pdfjs-setup-env.js: src/pdfjs-setup-env.js
-	cp $< $@
-build/pdfjs-worker-init.js: src/pdfjs-worker-init.js
+build/pdfjs-%.js: src/pdfjs-%.js
 	cp $< $@
 build/pdfjs: src/vendor/pdfjs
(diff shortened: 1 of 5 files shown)


robertknight:
An alternative approach here would be to revert https://github.com/hypothesis/browser-extension/pull/624, however I would really prefer to keep the stricter CSP policy in place.

PR #644, opened 2021-09-08
Update npm libraries before make commands

esanzgar:
I added the rule to update the npm libraries before running make
`test|typecheck|format|checkformatting` commands.


robertknight (approved):

robertknight:
I need to make a commit on the master branch to force a version bump, so I'm going to merge this.

PR #645, opened 2021-09-09
Support injecting client into VitalSource Bookshelf

robertknight:
**Depends on https://github.com/hypothesis/browser-extension/pull/663**

This PR adds support for loading the client into the VitalSource Bookshelf reader from within the browser extension context.

The initial use case is to make it easier for developers to test the VitalSource integration with various versions of the client (development, QA, production) and without having to go through the LMS app, which complicates the testing environment significantly.

Compared to a regular HTML document, the main difference with loading the client into bookshelf is that we need to target a specific frame which is not the top-level frame in the tab. The current implementation very specifically targets bookshelf, but we could make this more generic. In other words, provide infrastructure in the extension to teach it to inject the client into specific frames depending on the top-level URL in the tab. This might support other use cases in future.

### DIFF
--- a/src/background/chrome-api.js
+++ b/src/background/chrome-api.js
@@ -26,6 +26,15 @@ export function getChromeAPI(chrome = globalThis.chrome) {
    */
 
   /**
+   * Cache of Promise-ified APIs. This is used so that APIs which are looked up
+   * on-demand, eg. because they are optional and may not exist when `getChromeAPI`
+   * is first called, are only created once.
+   *
+   * @type {Map<Function, any>}
+   */
+  const cache = new Map();
+
+  /**
    * Convert an async callback-accepting Chrome API to a Promise-returning version.
    *
    * TypeScript may complain if the API has a Manifest V3-only overload that
@@ -36,9 +45,13 @@ export function getChromeAPI(chrome = globalThis.chrome) {
    * @param {(...args: [...Args, Callback<Result>]) => void} fn
    * @return {(...args: Args) => Promise<Result>}
    */
-  const promisify =
-    fn =>
-    (...args) => {
+  const promisify = fn => {
+    const cached = cache.get(fn);
+    if (cached) {
+      return cached;
+    }
+
+    return (...args) => {
       return new Promise((resolve, reject) => {
         fn(...args, (/** @type {Result} */ result) => {
           const lastError = chrome.runtime.lastError;
@@ -50,6 +63,7 @@ export function getChromeAPI(chrome = globalThis.chrome) {
         });
       });
     };
+  };
 
   /**
    * @template {any[]} Args
@@ -83,6 +97,10 @@ export function getChromeAPI(chrome = globalThis.chrome) {
       getURL: chrome.runtime.getURL,
     },
 
+    permissions: {
+      request: promisify(chrome.permissions.request),
+    },
+
     tabs: {
       create: promisify(chrome.tabs.create),
       get: promisifyAlt(chrome.tabs.get),
@@ -101,6 +119,20 @@ export function getChromeAPI(chrome = globalThis.chrome) {
       sync: {
         get: promisify(chrome.storage.sync.get.bind(chrome.storage.sync)),
       },
+    },
+
+    // APIs that require optional permissions.
+    //
+    // These are resolved on-demand because the `chrome.<namespace>` properties
+    // will not exist unless the extension has the required permissions.
+    //
+    // If a permission is revoked, the property remains accessible until the
+    // page/worker is reloaded, but calling APIs may fail.
+
+    get webNavigation() {
+      return {
+        getAllFrames: promisify(chrome.webNavigation.getAllFrames),
+      };
     },
   };
 }
--- a/src/background/sidebar-injector.js
+++ b/src/background/sidebar-injector.js
@@ -9,6 +9,7 @@ import {
 
 const CONTENT_TYPE_HTML = 'HTML';
 const CONTENT_TYPE_PDF = 'PDF';
+const CONTENT_TYPE_VITALSOURCE = 'VITALSOURCE';
 
 /** @param {Function} fn */
 function toIIFEString(fn) {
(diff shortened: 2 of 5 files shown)


robertknight:
I have refactored this to use the new mechanism for accessing extension APIs added in https://github.com/hypothesis/browser-extension/pull/663.

This simplifies using a new API (`chrome.webNavigation.getAllFrames`) in this PR. Some changes to `src/background/chrome-api.js` were required to deal with optional permissions.

robertknight:
You can test this by going to https://bookshelf.vitalsource.com/reader/books/9781400847402/epubcfi/6/22[%3Bvnd.vst.idref%3Dch2]!/4/2/5:0[%2C%202] (you might need a VitalSource account, but I believe this book is freely available) and activating the extension.

On master, the extension will inject the client into the top-level frame:

VitalSource inject master

On this branch, the extension will inject the client into the ebook reader frame:

VitalSource inject reader frame

You can't actually annotate the book content yet, because that requires the client changes in https://github.com/hypothesis/client/pull/3694 to be finished and shipped.

src/background/sidebar-injector.js
```diff
@@ -336,6 +350,81 @@ export function SidebarInjector() {
...
 
+  /**
+   * Find the frame within the VitalSource Bookshelf reader into which the
+   * Hypothesis client should be loaded.
+   *
+   * The frame hierarchy will look like:
+   *
+   * bookshelf.vitalsource.com (Main frame)
+   * |- jigsaw.vitalsource.com (Ebook reader. This is where we want to inject the client)
+   *     |- jigsaw.vitalsource.com (Content of current chapter)
```
lyzadanger, line 361: Thank you for this.

src/background/sidebar-injector.js
```diff
@@ -336,6 +350,81 @@ export function SidebarInjector() {
...
+    if (!frames) {
+      throw new Error('Could not list frames in tab');
+    }
+
+    return frames.find(frame => {
+      const frameURL = new URL(frame.url);
+      if (
+        frameURL.hostname !== 'jigsaw.vitalsource.com' ||
+        !frameURL.pathname.startsWith('/mosaic/wrapper.html')
+      ) {
```
lyzadanger, line 392: I sense a touch of long-term fragility here, but probably nothing you're not aware of already...

src/background/sidebar-injector.js
```diff
@@ -336,6 +350,81 @@ export function SidebarInjector() {
...
+        return null;
+      }
+
+      return frame;
+    });
+  }
+
+  /**
+   * @param {Tab} tab
+   * @param {object} config
```
lyzadanger, line 402: Should this be typed as `clientConfig`?

tests/background/chrome-api-test.js
```diff
@@ -87,4 +91,32 @@ describe('getChromeAPI', () => {
...
     assert.equal(error, fakeChrome.runtime.lastError);
   });
+
+  describe('APIs that require optional permissions', () => {
+    it('rejects if permission has not been granted', async () => {
+      const chromeAPI = getChromeAPI(fakeChrome);
+
+      let error;
+      try {
+        await chromeAPI.webNavigation.getAllFrames();
```
lyzadanger, line 101: This feels a little magical. Is it that the getter for `webNavigation` is not mocked, so it will behave in its "real" way here, which is to deny? Can you add some commenting?

tests/background/sidebar-injector-test.js
```diff
@@ -25,6 +25,25 @@ function createTestFrame() {
   return frame;
 }
 
+// Example of real VitalSource URLs for: 1) The top level frame, 2) The ebook reader frame, where we need
+// to inject the client and 3) the ebook chapter content frame.
```
lyzadanger, line 29: Thank you again for detailed commenting!

tests/background/sidebar-injector-test.js
```diff
@@ -76,10 +102,31 @@ describe('SidebarInjector', function () {
...
+          const allowed = request.permissions.every(perm =>
+            allowedPermissions.includes(perm)
+          );
+          if (allowed) {
+            request.permissions.forEach(perm => permissions.add(perm));
+          }
+          return allowed;
+        }),
+      },
+
```
lyzadanger, line 116: The mocking for things in the Chrome API in this test module is relatively involved; I don't know if there's any alternative...

tests/background/sidebar-injector-test.js
```diff
@@ -208,6 +255,70 @@ describe('SidebarInjector', function () {
       });
     });
 
+    describe('when viewing a VitalSource book', () => {
```
lyzadanger, line 258: This whole series of tests could be really complicated, but they're clear and easy to understand. Good job.

src/background/chrome-api.js
```diff
@@ -36,9 +45,13 @@ export function getChromeAPI(chrome = globalThis.chrome) {
...
-  const promisify =
-    fn =>
-    (...args) => {
+  const promisify = fn => {
+    const cached = cache.get(fn);
+    if (cached) {
+      return cached;
+    }
+
+    return (...args) => {
```
lyzadanger, line 54: This resolves a comment I left in #663 ... the explicit `return` here is easier to read. Thanks!

lyzadanger (approved):
From a code-review perspective, this looks good. I wasn't sure if you'd like any additional help with other kinds of testing?

src/background/sidebar-injector.js
robertknight: This class doesn't currently know anything about the configuration data that is injected, other than that it must be some JSON-serializable object.

tests/background/chrome-api-test.js
robertknight: Setting the `fakeChrome.webNavigation` property simulates the permission having been granted. I'll add a comment.

tests/background/sidebar-injector-test.js
robertknight: I guess the main thing that we could do is to centralize the mocks if this logic gets repeated across multiple tests. So far only this one class relies on this more complex faking.

robertknight:
> I wasn't sure if you'd like any additional help with other kinds of testing?

We discussed the manual testing steps in https://github.com/hypothesis/browser-extension/pull/645#issuecomment-924857642 on Slack. An issue came up that Lyza didn't have access to the book. For anyone else encountering this, an alternate URL is https://bookshelf.vitalsource.com/reader/books/BOOKSHELF-TUTORIAL/epubcfi/6/22[%3Bvnd.vst.idref%3Dvst-75af51a6-e0c3-480a-8c09-54bf7680703d]!/4.

PR #646, opened 2021-09-10
Modernize and add types to `promisify` utility

robertknight:
This PR is some cleanup and modernization of code that I'm touching while working on VitalSource integration (see https://github.com/hypothesis/browser-extension/pull/645).

Modernize and add types [1] to the `promisify` utility that we use to convert
callback-returning Chrome extension APIs to Promise-returning ones [2].

Adding these types triggered some conflicts in an `injectIntoRemoteDocument`
helper in `sidebar-injector.js` due to different branches of a function
returning incompatible types. Refactoring this function to async/await resolved
this issue by discarding unused information about the results (the caller of
`injectIntoRemoteDocument` only cares about when the operation has completed).

[1] This makes use of variadic tuple types. See
    https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html

[2] Chrome is migrating their APIs to return promises, so eventually we might not need this utility any more. From what I've seen the Promise-returning variants are only available after upgrading to Manifest V3 though, which entails many other changes.

### DIFF
--- a/src/background/util.js
+++ b/src/background/util.js
@@ -7,38 +7,40 @@ function getLastError() {
 }
 
 /**
- * Converts an async Chrome API into a function
- * which returns a promise.
+ * @template T
+ * @typedef {(result: T) => void} Callback
+ */
+
+/**
+ * Converts an async Chrome API which accepts a callback into a Promise-returning
+ * version.
  *
- * Usage:
- *   var apiFn = promisify(chrome.someModule.aFunction);
+ * @example
+ *   const apiFn = promisify(chrome.someModule.aFunction);
  *   apiFn(arg1, arg2)
- *     .then(function (result) { ...handle success  })
- *     .catch(function (err) { ...handle error })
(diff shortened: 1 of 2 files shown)


src/background/util.js
```diff
@@ -7,38 +7,40 @@ function getLastError() {
...
+ *     .then(result => { ... })
+ *     .catch(err => { ... })
  *
- * @param fn A Chrome API function whose last argument is a callback
- *           which is invoked with the result of the query. When this callback
- *           is invoked, the promise is rejected if chrome.extension.lastError
- *           is set or resolved with the first argument to the callback otherwise.
+ * @template {any[]} Args
+ * @template Result
+ * @param {(...args: [...Args, Callback<Result>]) => void} fn -
```
esanzgar, line 26: 👍🏼

esanzgar (approved):
Very nice

PR #647, opened 2021-09-10
Replace a query-string parsing utility with URLSearchParams

robertknight:


lyzadanger (approved):
The code looks good; let me know if there's any specific testing I could do to verify.

PR #654, opened 2021-09-16
Update the browser version support

esanzgar:
After the latest PDF.js version update, we require chrome >= 85.  See
https://hypothes-is.slack.com/archives/C1M8NH76X/p1631787431008100

Moved the browser list from .babelrc to package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "watchify": "^4.0.0",
     "web-ext": "^6.4.0"
   },
+  "browserslist": "chrome 85, firefox 75",
   "scripts": {
     "test": "karma start tests/karma.config.js --single-run",
     "typecheck": "tsc --build src/tsconfig.json"
--- a/.babelrc
+++ b/.babelrc
@@ -1,13 +1,3 @@
 {
-  "presets": [
-    [
-      "@babel/preset-env",
-      {
-        "targets": {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -46,6 +46,10 @@
     "watchify": "^4.0.0",
     "web-ext": "^6.4.0"
   },
+  "browserslist": {
+    "chrome": "85",
+    "firefox": "68"
```
robertknight, line 51: We can bump the Firefox requirement to v75 at the same time.

esanzgar: Done, thanks.

robertknight (approved):
LGTM

robertknight:
Something is failing on CI:

```
  ReferenceError: regeneratorRuntime is not defined
      at Suite.<anonymous> (/tmp/background/tab-state-test.js:157:7 <- /tmp/84a23033f0d469b186e8e9c33b4e5e84.browserify.js:21059:99)
```

Nothing should be referencing `regeneratorRuntime`. That's part of Babel's polyfill for async/await.

Can you check the compiled code and make sure it is actually respecting the browserslist config?

esanzgar:
I resolve it the issue. `browserslist` is a list not an object. I format it correctly and it is working.

PR #655, opened 2021-09-16
Simplify async code in SidebarInjector

robertknight:
Simplify async code in SidebarInjector by using async/await to replace
promise chains and `promisify` to convert some callback-accepting Chrome
APIs to Promise-returning functions.

### DIFF
--- a/src/background/sidebar-injector.js
+++ b/src/background/sidebar-injector.js
@@ -146,16 +146,14 @@ export function SidebarInjector() {
    *
    * @param {string} url
    */
-  function canInjectScript(url) {
-    let canInject;
+  async function canInjectScript(url) {
     if (isSupportedURL(url)) {
-      canInject = Promise.resolve(true);
+      return true;
     } else if (isFileURL(url)) {
-      canInject = chromeAPI.extension.isAllowedFileSchemeAccess();
+      return chromeAPI.extension.isAllowedFileSchemeAccess();
     } else {
-      canInject = Promise.resolve(false);
+      return false;
     }
-    return canInject;
   }
 
   /**
@@ -180,40 +178,36 @@ export function SidebarInjector() {
   }
 
   /** @param {Tab} tab */
-  function detectTabContentType(tab) {
+  async function detectTabContentType(tab) {
     if (isPDFViewerURL(tab.url)) {
-      return Promise.resolve(CONTENT_TYPE_PDF);
+      return CONTENT_TYPE_PDF;
     }
 
     if (isVitalSourceURL(tab.url)) {
-      return Promise.resolve(CONTENT_TYPE_VITALSOURCE);
+      return CONTENT_TYPE_VITALSOURCE;
     }
 
-    return canInjectScript(tab.url).then(function (canInject) {
-      if (canInject) {
-        return chromeAPI.tabs
-          .executeScript(tab.id, {
-            code: toIIFEString(detectContentType),
-          })
-          .then(function (frameResults) {
-            const result = extractContentScriptResult(frameResults);
-            if (result) {
-              return result.type;
-            } else {
-              // If the content script threw an exception,
-              // frameResults may be null or undefined.
-              //
-              // In that case, fall back to guessing based on the
-              // tab URL
-              return guessContentTypeFromURL(tab.url);
-            }
-          });
+    const canInject = await canInjectScript(tab.url);
+    if (canInject) {
+      const frameResults = await chromeAPI.tabs.executeScript(tab.id, {
+        code: toIIFEString(detectContentType),
+      });
+      const result = extractContentScriptResult(frameResults);
+      if (result) {
+        return result.type;
       } else {
-        // We cannot inject a content script in order to determine the
-        // file type, so fall back to a URL-based mechanism
-        return Promise.resolve(guessContentTypeFromURL(tab.url));
+        // If the content script threw an exception,
+        // frameResults may be null or undefined.
+        //
+        // In that case, fall back to guessing based on the
+        // tab URL
+        return guessContentTypeFromURL(tab.url);
       }
-    });
+    } else {
+      // We cannot inject a content script in order to determine the
+      // file type, so fall back to a URL-based mechanism
+      return guessContentTypeFromURL(tab.url);
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/background/sidebar-injector.js
```diff
@@ -187,34 +187,32 @@ export default function SidebarInjector(
...
+      const result = extractContentScriptResult(frameResults);
+      if (result) {
+        return result.type;
       } else {
-        // We cannot inject a content script in order to determine the
-        // file type, so fall back to a URL-based mechanism
-        return Promise.resolve(guessContentTypeFromURL(tab.url));
+        // If the content script threw an exception,
+        // frameResults may be null or undefined.
+        //
```
esanzgar, line 206: If the content script really throws an exception, then `promisify` rejects the promise and we should never reach this point.

I think it would be better to add a type to `result` and remove this comment.
```
      /** @type {{type:string}|null|undefined} - null|undefined if unable to get the contentType */
```

src/background/sidebar-injector.js
```diff
@@ -323,29 +314,23 @@ export default function SidebarInjector(
...
-      const originalURL = parsedURL.searchParams.get('file');
-      if (!originalURL) {
-        throw new Error('Failed to extract original URL from ' + tab.url);
-      }
-      let hash = parsedURL.hash;
+  async function removeFromPDF(tab) {
+    const parsedURL = new URL(tab.url);
+    const originalURL = parsedURL.searchParams.get('file');
+    if (!originalURL) {
+      throw new Error('Failed to extract original URL from ' + tab.url);
```
esanzgar, line 321:
```suggestion
      throw new Error(`Failed to extract original URL from ${tab.url}`);
```

esanzgar (commented):
I have a question about a comment that seems to indicate that the (remote) execution of one script throws an exception. The code doesn't reflect that case, so I think the language is misleading. I suggest to remove the comment and replace it with a  type.

src/background/sidebar-injector.js
esanzgar: Could we add the type to the  `frameResults` and modify `extractContentScriptResult` to preserve the type?

```
      /** @type {Array<{type:string}|undefined>} - [undefined] if unable to get the contentType */
      const frameResults = await executeScript(tab.id, {
....
```

```
/**
 * Extract the value returned by a content script injected via
 * chrome.tabs.executeScript() into the main frame of a page.
 *
 * executeScript() returns an array of results, one per frame which the script
 * was injected into.
 *
 * See https://developer.chrome.com/extensions/tabs#method-executeScript
 * @template T
 * @param {(Array<T>|T)?} result
 */
function extractContentScriptResult(result) {
  if (Array.isArray(result) && result.length > 0) {
    return /** @type {T} */ (result[0]);
  } else if (typeof result === 'object') {
    // Firefox currently returns an object instead of
    // an array from executeScript()
    return /** @type {T} */ (result);
  } else {
    return null;
  }
}
```

robertknight: > If the content script really throws an exception, then promisify rejects the promise and we should never reach this point.

What makes you think the promise will reject if the content script throws?

In my testing the comment is still accurate. Try the following:

1. Apply this diff and rebuild the extension
2. Activate the Hypothesis extension on a simple HTML page
3. Look at the console logs for the extension's background page (linked from the extension's entry in chrome://extensions). You should see that the value of `frameResults` is `[null]`.

```
diff --git a/src/background/detect-content-type.js b/src/background/detect-content-type.js
index c723f16..0748132 100644
--- a/src/background/detect-content-type.js
+++ b/src/background/detect-content-type.js
@@ -14,6 +14,8 @@
  */
 /* istanbul ignore next */
 export default function detectContentType(document_ = document) {
+  throw new Error('Exception from detectContentType');
+
   function detectChromePDFViewer() {
     // When viewing a PDF in Chrome, the viewer consists of a top-level
     // document with an <embed> tag, which in turn instantiates an inner HTML
diff --git a/src/background/sidebar-injector.js b/src/background/sidebar-injector.js
index f8dba88..6f978f8 100644
--- a/src/background/sidebar-injector.js
+++ b/src/background/sidebar-injector.js
@@ -197,6 +197,7 @@ export default function SidebarInjector(
       const frameResults = await executeScript(tab.id, {
         code: toIIFEString(detectContentType),
       });
+      console.log('detectContentType script results', frameResults);
       const result = extractContentScriptResult(frameResults);
       if (result) {
         return result.type;
```

robertknight: > Could we add the type to the frameResults and modify extractContentScriptResult to preserve the type?

In my testing with the current version of Firefox, it seems like the comment about Firefox returning an object rather than an array is no longer valid. On that basis I think we can just remove `extractContentScriptResult` entirely and rely on either `executeScript` throwing or returning an array.

robertknight:
@esanzgar - I made the string interpolation change. Regarding the `executeScript()` call, the comment about the result possibly being `null`, if the content script throws an exception, is still valid in my testing. I did discover that the deviation between Firefox and Chrome's behavior no longer seems to apply, so I think the `extractContentScriptResult` helper can just be removed. I'd like to make that a separate change however.

src/background/sidebar-injector.js
esanzgar: You are correct, if the content script throws the promise is not rejected.

The `chromeTabs.executeScript` is promisified. The following lines of code in the `promisify` function lead me to believe that the promise could be rejected: https://github.com/hypothesis/browser-extension/blob/cb63a86661e4e81a17d3fbaae42c43e9bcc0571f/src/background/chrome-api.js#L55-L61

I have found that's not the case and I will address that separately.

esanzgar (approved):
Thanks for clarifying my question.

src/background/sidebar-injector.js
robertknight: > The following lines of code in the promisify function lead me to believe that the promise could be rejected:

You're right though. The `chrome.tabs.executeScript` call _could_ fail, which would result in the Promise being rejected. It just happens that this is only the only way that a failure can happen in this case, and it is not how errors in the content script get handled.

esanzgar: Oddly, the `if (lastError)` path *is* taken when accessed a chrome web store page, like: https://chrome.google.com/webstore/detail/hypothesis-web-pdf-annota/bjfhmglciegochdpefhhlphglcehbmek

The above page doesn't allow script execution.

The discrepancy between the two scenarios is a bit puzzling.

PR #656, opened 2021-09-16
Convert default exports to named exports

robertknight:
Per current coding conventions, use named rather than default exports.


esanzgar (approved):
I like this general modernisation. Thanks!

PR #657, opened 2021-09-16
Add remaining missing types and disable `noImplicitAny`

robertknight:
Add remaining missing types so that the code typechecks with `noImplicitAny`
enabled in tsconfig.json.

### DIFF
--- a/src/background/browser-action.js
+++ b/src/background/browser-action.js
@@ -17,8 +17,12 @@ const icons = {
   },
 };
 
-// themes to apply to the toolbar icon badge depending on the type of
-// build. Production builds use the default color and no text
+/**
+ * Themes to apply to the toolbar icon badge depending on the type of
+ * build. Production builds use the default color and no text
+ *
+ * @type {Record<string, { defaultText: string, color: string }>}
+ */
 const badgeThemes = {
   dev: {
     defaultText: 'DEV',
@@ -29,11 +33,6 @@ const badgeThemes = {
     color: '#EDA061', // Porche orange-pink
   },
 };
-
-// Fake localization function.
-function _(str) {
-  return str;
-}
 
 /**
  * Controls the display of the browser action button setting the icon, title
@@ -85,11 +84,9 @@ export function BrowserAction(chromeBrowserAction) {
 
       let countLabel;
       if (state.annotationCount === 1) {
-        countLabel = _("There's 1 annotation on this page");
+        countLabel = "There's 1 annotation on this page";
       } else {
-        countLabel = _(
-          'There are ' + totalString + ' annotations on this page'
-        );
+        countLabel = `There are ${totalString} annotations on this page`;
       }
 
       title = countLabel;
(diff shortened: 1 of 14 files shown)


src/background/hypothesis-chrome-extension.js
```diff
@@ -150,20 +154,24 @@ export default function HypothesisChromeExtension({
   // exposed for use by tests
   this._onTabStateChange = onTabStateChange;
 
+  /** @param {chrome.tabs.Tab} tab */
   function onBrowserActionClicked(tab) {
-    const tabError = state.getState(tab.id).error;
+    const tabId = /** @type {number} */ (tab.id);
```
robertknight, line 159: I am assuming in several places in this PR that tabs have an ID and URL. We've never seen any reports of errors due to this not being the case, though I couldn't completely guarantee that it has never happened.

src/background/settings.js
```diff
@@ -1,7 +1,16 @@
+/**
+ * Incomplete type for settings in the `settings.json` file.
```
robertknight, line 2: The intent is that we'll fill this out with additional fields as we need to.

src/background/browser-action.js
```diff
@@ -17,8 +17,12 @@ const icons = {
   },
 };
 
-// themes to apply to the toolbar icon badge depending on the type of
-// build. Production builds use the default color and no text
+/**
+ * Themes to apply to the toolbar icon badge depending on the type of
+ * build. Production builds use the default color and no text
+ *
+ * @type {Record<string, { defaultText: string, color: string }>}
```
esanzgar, line 24:
```suggestion
 * @type {Record<string, {defaultText: string, color: string}>}
```

src/background/tab-state.js
```diff
@@ -77,14 +79,14 @@ export default function TabState(initialState, onchange) {
...
-   *                   The provided state will be merged with the default
-   *                   state for a tab.
+   * @param {Record<number, Partial<State>>} newState -
+   *   Map of tab ID to state. The state is merged with the default state for a tab.
    */
   this.load = function (newState) {
     /** @type {TabStateMap} */
     const newCurrentState = {};
-    Object.keys(newState).forEach(function (tabId) {
+    Object.keys(newState).forEach(function (tabIdStr) {
```
esanzgar, line 88:
```suggestion
    Object.keys(newState).forEach(tabIdStr => {
```

As part of this modernisation, we could run `eslint` to automatically convert to arrow functions, when possible.

esanzgar (approved):
LGTM

src/background/browser-action.js
robertknight: There aren't a ton of examples but we do tend to include a space before and after parens in object types. The general rule of thumb I follow is to format types as Prettier would (in a TypeScript file, unfortunately it doesn't format JSDoc types).

src/background/tab-state.js
robertknight: Yes. This was on my list to do at some point.

PR #659, opened 2021-09-20
Automatic type substitution of Object for object

esanzgar:
For the past few months, we have started to use the `object` type instead of `Object`. Both are almost interchangeable, but we prefer `object` for simplicity.

This PR substitutes all `Object` to `object` and makes the code more consistent.

I used these commands (Mac version):

```
% ag -0 -l '@.+Object' | xargs -0 sed -ri '' -e '/@/s/(.*{.*)Object(.*})/\1object\2/'
```

### DIFF
--- a/src/background/direct-link-query.js
+++ b/src/background/direct-link-query.js
@@ -4,7 +4,7 @@
  *
  * See https://h-client.readthedocs.io/en/latest/publishers/config/#config-settings
  *
- * @typedef {Object} Query
+ * @typedef Query
  * @property {string} [annotations] - ID of the direct-linked annotation
  * @property {string} [query] - Filter query from the sidebar
  * @property {string} [group] - ID of the direct-linked group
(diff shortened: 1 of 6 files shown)


src/background/direct-link-query.js
```diff
@@ -4,7 +4,7 @@
  *
  * See https://h-client.readthedocs.io/en/latest/publishers/config/#config-settings
  *
- * @typedef {Object} Query
+ * @typedef {object} Query
```
robertknight, line 7: Current conventions for typedefs are to omit the `{object}` part if there are `@prop` definitions below, because it can be inferred in that case.

robertknight (approved):
LGTM. I think the main reason to prefer `object` is that it is a standard TypeScript type and the meaning is more obvious. Looking at the implementation, it seems that the meaning in a JS file depends on whether `noImplicitAny` is enabled. If it is, `object`/`Object` are aliases of `any`, otherwise they mean "some non primitive type".

src/background/direct-link-query.js
esanzgar: 👍🏼

PR #667, opened 2021-09-23
Promisify can't reject Promise

esanzgar:
Exceptions in a promisified function can only happened in the execution
of the callback. In our cases, the callback is very simple and will
never raised an exception. Hence the rejection path can be removed.

Related to https://github.com/hypothesis/browser-extension/pull/655#pullrequestreview-757394610.

### DIFF
--- a/src/background/chrome-api.js
+++ b/src/background/chrome-api.js
@@ -52,15 +52,8 @@ export function getChromeAPI(chrome = globalThis.chrome) {
     }
 
     return (...args) => {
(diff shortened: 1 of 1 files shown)


esanzgar:
After discussion with Robert, the rejection path is necessary. He demonstrated it with this example:

>  Go to https://chrome.google.com/webstore/detail/hypothesis-web-pdf-annota/bjfhmglciegochdpefhhlphglcehbmek/related?hl=en and try to activate the extension.

PR #668, opened 2021-09-23
Eliminate side effects from `install.js` import in tests

robertknight:
This simplifies the tests by allowing `install.js` to be imported
normally, as the `chrome` fake can be injected before `init` is run.


esanzgar (approved):
👍🏼

PR #673, opened 2021-09-27
Improve type for extractContentScriptResult

esanzgar:
1. Robert and I independently have confirmed that the newer versions of
   Firefox return an Array from `chrome.tab.executeScript`.

2. Added a sentence to clarify when the value of `null` can be found in
   the array of results.

3. Modify the type for the `extractContentScriptResult` so that the type
   is preserved.


src/background/sidebar-injector.js
```diff
@@ -189,6 +187,7 @@ export function SidebarInjector() {
 
     const canInject = await canInjectScript(tab.url);
     if (canInject) {
+      /** @type {{type: string}[]}} */
       const frameResults = await chromeAPI.tabs.executeScript(tab.id, {
         code: toIIFEString(detectContentType),
       });
```
esanzgar, line 193: The type of `result` variable can be correctly inferred.
 
Screenshot 2021-09-27 at 15 22 56

esanzgar:
Related to the `extractContentScriptResult`, I have check the following lines:

https://github.com/hypothesis/browser-extension/blob/991759c5814d6e41c0bbcb8b19a2fce82074427d/src/background/sidebar-injector.js#L277-L286

It is not clear if the injector ever return an object with the `installedURL` property. At least, I couldn't make it happened.

src/background/sidebar-injector.js
```diff
@@ -39,19 +39,17 @@ function addJSONScriptTag(name, content) {
  * chrome.tabs.executeScript() into the main frame of a page.
  *
  * executeScript() returns an array of results, one per frame which the script
- * was injected into.
+ * was injected into. If the frame fails to execute the script it returns a value
+ * of `null`.
```
robertknight, line 43:
```suggestion
 * was injected into. If a frame fails to execute the script the array entry will
 * have a value of `null`.
```

I reworded this to clarify that the `null` refers to entries in the `result` array, not `result` itself.

src/background/sidebar-injector.js
```diff
@@ -39,19 +39,17 @@ function addJSONScriptTag(name, content) {
...
 function extractContentScriptResult(result) {
-  if (Array.isArray(result) && result.length > 0) {
+  if (result.length > 0) {
     return result[0];
-  } else if (typeof result === 'object') {
-    // Firefox currently returns an object instead of
-    // an array from executeScript()
-    return result;
   } else {
     return null;
```
robertknight, line 54: I'm not sure if this branch can ever actually be hit in practice, because every tab presumably has at least one frame.

src/background/sidebar-injector.js
```diff
@@ -189,6 +187,7 @@ export function SidebarInjector() {
 
     const canInject = await canInjectScript(tab.url);
     if (canInject) {
+      /** @type {{type: string}[]}} */
```
robertknight, line 190:
```suggestion
      /** @type {Array<{ type: string } | null>} */
```

Entries within this result could be `null`. I changed from `T[]` to `Array<T>` syntax here because this introduced a union.

robertknight (commented):
I added a couple of minor corrections/suggestions.

PR #674, opened 2021-09-28
Replace Browserify with Rollup

robertknight:
Replace Browserify with a more modern bundler that understands ES modules
natively and has better ecosystem support.

Apart from the bundler change, there is also a change to how the code is bundled
for tests. Instead of using a Karma framework, we create a bundle ourselves from
the test files and then instruct Karma to load the pre-bundled code. This
eliminates the need for developers to understand a glue dependency
(karma-browserify) and also makes it easier to inspect the generated test
bundle, since it is output to the build directory before the tests run.

## Why Rollup?

On a front-end sync last week we identified various pain points caused by the current bundler that our projects use (Browserify). The most immediate ones were those reported by Lyza while trying to integrate react-intl in the LMS app. This was a lack of built-in support for global aliasing (eg. for React => Preact usage) as well as general support in the ecosystem. I also identified the lack of native support for ES modules an issue going forwards.

There several viable candidates for a replacement for Browserify. These range from more minimalist bundlers which feel like modern successors to Browserify (eg. Rollup) to more kitchen-sync bundlers like Webpack and Parcel which are designed more as complete solutions for building SPA apps. I have not tried them all extensively and I expect we could make most of them work. After we have done the work to prepare our projects for use with an ES-native bundler, I also expect that switching between bundlers would be quite manageable if found a clear need. I'm suggesting Rollup here because I think it meets our requirements, I was able to get it to work reasonably easily for our most complex app (the client) and it was recommended by some folks I trust (devs working on Preact and Chrome devtools).

Some pros and cons from my experimentation so far:

Pros:

- Has native support for ES modules, including features like dynamic import
- Has most of the functionality we need available in the bundler itself or as a plugin maintained by the core team
- Is widely used in the current JS ecosystem, such that there is good documentation on how to use it with various projects
- Has good CLI and JavaScript APIs, so easy to use either from scripts (eg. in package.json) or gulpfile (for more complex use cases)
- Was recommended by members of the core Preact team when I asked for their input
- Structure of generated bundles is easy to follow
- I was able to create working builds of the client and browser extension without too much difficulty
- Has good documentation on how to create plugins, in case we need to add more advanced features
- Does tree shaking (eg. so parts of the @hypothesis/frontend-shared project that are unused by a particular app will get stripped out)

Cons:

- Not as fast as some alternatives (eg. bundlers like esbuild which are written in non-JS languages) and doesn't have built-in disk caching. Current performance is about the same as with Browserify.


.eslintignore
```diff
@@ -2,5 +2,8 @@ build/**
 dist/*
 **/vendor/**/*.js
 
+# TODO - Lint these files
+rollup*.config.js
```
robertknight, line 6: ESLint gets confused by the fact that Rollup config files are ES modules. It isn't super-important to lint these, but we can also probably fix this by telling ESLint to assume all files in the repo are ES modules by default.

Makefile
```diff
@@ -32,7 +31,7 @@ dev: node_modules/.uptodate
 .PHONY: clean
 clean:
 	rm -f node_modules/.uptodate
-	rm -rf build/* build/.settings.json build/.*.deps
```
robertknight, line 35: I removed all the logic related to building the `.deps` file because IMO it doesn't add a lot of value. We can just rebuild the bundle whenever any JS file in `src/` changes or whenever the yarn lockfile changes.

rollup-tests.config.js
```diff
@@ -0,0 +1,39 @@
+import * as glob from 'glob';
```
robertknight, line 1: We might want to put the rollup configs in a separate directory if there are lot of them. In this project there are only two however.

rollup-tests.config.js
```diff
@@ -0,0 +1,39 @@
...
+import { babel } from '@rollup/plugin-babel';
+import commonjs from '@rollup/plugin-commonjs';
+import json from '@rollup/plugin-json';
+import multi from '@rollup/plugin-multi-entry';
+import { nodeResolve } from '@rollup/plugin-node-resolve';
```
robertknight, line 6: This is a pretty standard set of plugins. I think most of them speak for themselves.

- @rollup/plugin-commonjs allows the use of CommonJS-based dependencies. This can be removed if all dependencies are ES module based
- @rollup/plugin-multi-entry supports bundles with multiple entry points. This is mainly useful for tests
- @rollup/plugin-node-resolve makes `import` statements resolve module specifies the same way as Node (and Browserify) do, which is different than browsers (eg. by not requiring the `.js` extension and by doing lookups in node_modules)

tests/background/hypothesis-chrome-extension-test.js
```diff
@@ -55,7 +55,7 @@ describe('HypothesisChromeExtension', function () {
   }
 
   beforeEach(function () {
-    global.chrome = { runtime: chromeRuntime };
+    globalThis.chrome = { runtime: chromeRuntime };
```
robertknight, line 58: See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis.

esanzgar (approved):
The bundle is smaller and it is created faster. Great job.

My only consideration is that the test bundle may be shipped with the extension unnecessarily.

robertknight:
> My only consideration is that the test bundle may be shipped with the extension unnecessarily.

Good catch! I've added a `make clean` step before the `make build` calls in Jenkinsfile.

PR #696, opened 2021-10-28
Move settings data into the main extension bundle

robertknight:
The settings data for the extension is currently loaded into the background page as a separate `settings-data.js` script before the main extension bundle. In Manifest V3 the background code runs in a Service Worker which has to be specified as a single JS script ( although it can be an ES module which imports other scripts).

To expose the settings to the background app in a way which works in both Manifest V2 and Manifest V3, this PR incorporates them directly into the main bundle. This works as follows:

 1. make runs `tools/settings.js` to generate `build/settings.json` based on
    the input specified via `SETTINGS_FILE=<path>`.

 2. `src/background/settings.js` now imports `build/settings.json` directly,
    using Rollup's JSON plugin to translate it to an ES module.
    
----

**Testing:**

Aside from checking that the extension builds and runs as usual, run `make build SETTINGS_FILE=settings/<filename>` for different files in the `settings/` directory and check that the values from the input file appear in the extension bundle. Search for `var oauthClientId` in `build/extension.bundle.js` to find the embedded settings.


rollup-tests.config.js
```diff
@@ -15,6 +16,14 @@ export default {
   },
   treeshake: false, // Disabled for build performance
   plugins: [
+    alias({
+      entries: [
+        {
+          find: '../../build/settings.json',
+          replacement: '../../tests/settings.json',
```
robertknight, line 23: In tests the `settings.json` import is re-pointed at a static `settings.json` file in the `tests/` dir.

src/background/settings.js
```diff
@@ -4,9 +4,13 @@
  * @typedef Settings
  * @prop {string} apiUrl
  * @prop {string} buildType
+ * @prop {{ dsn: string, release: string}} [raven]
```
esanzgar, line 7:
```suggestion
 * @prop {{dsn: string, release: string}} [raven]
```

src/background/settings.js
```diff
@@ -4,9 +4,13 @@
...
  * @prop {string} serviceUrl
  */
 
+// @ts-ignore - Ignore missing module if file has not been generated.
```
esanzgar, line 11: The JSON file is always present. The reason that this raise a warning is that TS doesn't understand importing JSON by default. It can be unable with:

```
"resolveJsonModule": true
```

I think it is worth considering adding this option.

One nice thing about that is that the types are defined automatically:

Screenshot 2021-11-04 at 13 19 29

src/background/settings.js
```diff
@@ -23,4 +27,4 @@ function normalizeSettings(settings) {
 /**
  * Returns the configuration object for the Chrome extension
  */
-export default normalizeSettings(/** @type {any} */ (window).EXTENSION_CONFIG);
+export default normalizeSettings(settings);
```
esanzgar, line 30: Interestingly, this could be one liner: `export default { ...settings, apiUrl: settings.apiUrl.replace(/\/$/, '') };`

esanzgar (approved):
This is much nicer, thank you very much!

src/background/index.js
```diff
@@ -1,10 +1,8 @@
 import * as raven from './raven';
-
-// @ts-expect-error - `EXTENSION_CONFIG` is missing from types
-if (window.EXTENSION_CONFIG.raven) {
-  // @ts-expect-error - `EXTENSION_CONFIG` is missing from types
-  raven.init(window.EXTENSION_CONFIG.raven);
-}
-
+import settings from './settings';
```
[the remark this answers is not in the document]
robertknight: This is true, although you have to be careful here since some of the imports have side effects so order may matter. In this case it was simple to rework the imports so that none of them have side effects, which allows them to be ordered as normal.

src/background/settings.js
robertknight: > The JSON file is always present. 

That's not true if the build hasn't run yet.

Thanks for the tip about `resolveJsonModule`. I have enabled it and modified the `make lint` command so it has a dependency on `build/settings.json` and also added a note for developers about why the file may see an error in their editor.

src/background/settings.js
robertknight: Indeed. I've made this change with one slight modification - the object needs to be cast to the `Settings` type to avoid errors in other modules about optional settings which happen not to exist in whatever settings file you most recently used.

PR #697, opened 2021-10-28
Remove remaining references to `window` global

robertknight:
In preparation for the Manifest V3 migration, remove remaining explicit
references to the `window` global.

 - Remove unused `browser-name.js` module
 - Replace `window` reference with `globalThis` in raven.js. The
   extension's crash reporting is due for an overhaul as raven-js is now
   obsolete, but this is a minimal change.

There is a remaining use of `localStorage` that will need to be replaced, but that will be done separately.


src/background/raven.js
```diff
@@ -115,7 +115,7 @@ export function report(error, when, context) {
  * automatically, in which case this code can simply be removed.
  */
 function installUnhandledPromiseErrorHandler() {
-  window.addEventListener('unhandledrejection', function (event) {
+  globalThis.addEventListener('unhandledrejection', function (event) {
```
robertknight, line 118: To recap, `globalThis` refers to `window` in a regular web page, but whatever the "global object" is in other environments. In Manifest V3, this will be a Service Worker.

lyzadanger (approved):
Coolio.

PR #705, opened 2021-11-04
Correct regex pattern in the replace

esanzgar:
I believe the intention is to strip out the last `/` character, if any
(as the comment say).


robertknight (approved):
LGTM 🤦‍♂️

PR #775, opened 2022-01-26
Format JS files in tools/ folder

robertknight:
Change the Prettier config to format all JS files except for ignored
directories. This ensures that all files in the root dir and tools/ are
formatted.


lyzadanger (approved):
Not so controversial, this one :)

PR #776, opened 2022-01-26
Remove `websocketUrl` setting

robertknight:
The client now obtains this URL from the `/api/links` endpoint.  See https://github.com/hypothesis/h/pull/7254 and
https://github.com/hypothesis/client/pull/4148.

This should not be merged until https://github.com/hypothesis/client/pull/4148 is deployed. (_Update: This is now done._)


lyzadanger (approved):
This is a good direction. Yay!

PR #797, opened 2022-02-15
Disable extension on LMS assignments

esanzgar:
LMS assignments can be opened on a new window. When that option is used
the URL of the window has an origin `[*]lms.hypothesis.is`.

image

If the the document's URL matches `[*]lms.hypothes.is` we disable the
browser extension and show a warning message (on second click).

Screenshot 2022-02-14 at 17 22 18

Closes https://github.com/hypothesis/product-backlog/issues/1310

### DIFF
--- a/src/background/errors.js
+++ b/src/background/errors.js
@@ -32,6 +32,8 @@ const IGNORED_ERRORS = [
   // Attempts to access pages for which Chrome does not allow scripting
   /Cannot access contents of.*/,
   /The extensions gallery cannot be scripted/,
+  // The extension is blocked on LMSs
+  /Hypothesis extension can't be used on learning platform assignments/,
 ];
 
 /**
--- a/src/background/sidebar-injector.js
+++ b/src/background/sidebar-injector.js
@@ -2,6 +2,7 @@ import { chromeAPI, executeFunction, executeScript } from './chrome-api';
 import { detectContentType } from './detect-content-type';
 import {
   AlreadyInjectedError,
+  BlockedSiteError,
   LocalFileError,
   NoFileAccessError,
   RestrictedProtocolError,
@@ -10,6 +11,7 @@ import {
 const CONTENT_TYPE_HTML = 'HTML';
 const CONTENT_TYPE_PDF = 'PDF';
 const CONTENT_TYPE_VITALSOURCE = 'VITALSOURCE';
+const CONTENT_TYPE_LMS = 'LMS';
 
 /**
  * @param {object} config
@@ -143,6 +145,12 @@ export function SidebarInjector() {
     return url.startsWith('https://bookshelf.vitalsource.com/');
   }
 
+  /** @param {string} url */
+  function isLMSAssignmentURL(url) {
+    const { origin } = new URL(url);
+    return origin.includes('lms.hypothes.is');
+  }
+
   /** @param {Tab} tab */
   async function detectTabContentType(tab) {
     if (isPDFViewerURL(tab.url)) {
@@ -151,6 +159,10 @@ export function SidebarInjector() {
 
     if (isVitalSourceURL(tab.url)) {
       return CONTENT_TYPE_VITALSOURCE;
+    }
+
+    if (isLMSAssignmentURL(tab.url)) {
+      return CONTENT_TYPE_LMS;
     }
 
     const canInject = await canInjectScript(tab.url);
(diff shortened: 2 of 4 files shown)


src/background/errors.js
```diff
@@ -32,6 +32,8 @@ const IGNORED_ERRORS = [
   // Attempts to access pages for which Chrome does not allow scripting
   /Cannot access contents of.*/,
   /The extensions gallery cannot be scripted/,
+  // The extension is blocked on LMSs
```
robertknight, line 35: We need to be clear that the extension is only disabled in Hypothesis LMS assignments to avoid confusion for students and teachers between the extension and the embedded Hypothesis instance. The user can still use the extension on other pages hosted in the LMS itself.

src/background/sidebar-injector.js
```diff
@@ -143,6 +145,12 @@ export function SidebarInjector() {
     return url.startsWith('https://bookshelf.vitalsource.com/');
   }
 
+  /** @param {string} url */
+  function isLMSAssignmentURL(url) {
+    const { origin } = new URL(url);
+    return origin.includes('lms.hypothes.is');
```
robertknight, line 151: This won't work for the Canada LMS, whose URL is `lms.ca.hypothes.is`. I would suggest to do something like:

1. Check that the TLD is `hypothes.is` (ie. host ends with `.hypothes.is`)
2. Check for the token `lms` in the subdomains (eg. by tokenizing into alphanumeric tokens)

src/background/sidebar-injector.js
```diff
@@ -239,6 +251,10 @@ export function SidebarInjector() {
       await injectIntoPDF(tab);
     } else if (type === CONTENT_TYPE_VITALSOURCE) {
       await injectIntoVitalSourceReader(tab, config);
+    } else if (type === CONTENT_TYPE_LMS) {
+      throw new BlockedSiteError(
+        "Hypothesis extension can't be used on learning platform assignments"
```
robertknight, line 256:
```suggestion
        "Hypothesis extension can't be used on Hypothesis LMS assignments"
```

We need to be clear that this is specifically about Hypothesis LMS assignments. The user might not know what "LMS" means, but we do use the term extensively in our docs so it is easy to search for. Searching Google for "Hypothesis LMS" turns up https://web.hypothes.is/education/lms/.

src/background/sidebar-injector.js
esanzgar, line 151: If via in the LMS context changes URL, we will need to change this.

robertknight (changes requested):
My original idea was that the extension would just be deactivated rather than blocked entirely on LMS assignments. However I think disabling the extension entirely will work just as well. I see two issues that need to be resolved:

1. The domain check won't work for Canada's LMS environment
2. Code comments and error messages shown to the user need to make it clear that only Hypothesis LMS assignments are blocked. The user can still use the extension elsewhere on the LMS platform. For example there might be course notes they would want to annotate.

src/background/sidebar-injector.js
```diff
@@ -239,6 +251,10 @@ export function SidebarInjector() {
       await injectIntoPDF(tab);
     } else if (type === CONTENT_TYPE_VITALSOURCE) {
       await injectIntoVitalSourceReader(tab, config);
+    } else if (type === CONTENT_TYPE_LMS) {
+      throw new BlockedSiteError(
```
robertknight, line 255: Can you add a comment here to indicate why this site is blocked, as it might not be obvious to future developers.

esanzgar:
I included your suggestion on a second commit. Please, let me know if you are happy with the changes.

src/background/sidebar-injector.js
```diff
@@ -143,6 +145,12 @@ export function SidebarInjector() {
     return url.startsWith('https://bookshelf.vitalsource.com/');
   }
 
+  /** @param {string} url */
+  function isLMSAssignmentURL(url) {
+    const { origin } = new URL(url);
+    return origin.includes('hypothes.is');
```
robertknight, line 151: This will match _any_ site hosted on the `hypothes.is` domain or subdomains! I was thinking of something like:

```js
// Matches origins like `lms.hypothes.is`, `qa-lms.hypothes.is`, `lms.ca.hypothes.is`.
origin.match(/\blms\b/) && origin.endsWith('.hypothes.is')
```

robertknight (changes requested):
The exception wording change looks good. The revised origin match is too broad though. See suggestion for what I had in mind.

src/background/sidebar-injector.js
esanzgar: What do you think to do it on a single regex? `/\blms\b\.*\.hypothes\.is$/.test(origin)`

robertknight (approved):
LGTM. I leave it up to you whether to change to a single regex or leave as-is.

PR #798, opened 2022-02-16
Add `--grep` and `--watch` options to the test command

esanzgar:
Similar to the client and LMS projects, I added support for the `--grep`
and `--watch` options for testing.

Because the browser-extension codebase is small it doesn't take much
time to bundle the files, so the `--grep` option is not super
needed. However, the `--watch` option is a useful one.

In any case, for consistency and developer experience, I think it is
better to have these two options available.

I copy-pasted the `runTest` function from the
`@hypothesis/frontend-build`. I didn't use this package directly because
it has a number of peer dependencies (sass, tailwindcss, etc.) to are
irrelevant to this project.

The test command can be invoke in a number of ways:
* make test ARGS="--watch --grep injector"
* gulp test --watch --grep injector
* yarn gulp test --watch --grep injector


robertknight:
> I didn't use this package directly because it has a number of peer dependencies (sass, tailwindcss, etc.) to are irrelevant to this project.

It might be possible to resolve this by using yarn aliases as dummy resolutions of these dependencies. Alternatively we could consider splitting up the package into separate ones.

robertknight:
I think I'd prefer to avoid copying this volume of code from the `@hypothesis/frontend-build` package if possible. This would require coming up with a solution regarding the unwanted Tailwind and SASS dependencies. This is not a major pain point for our current work and we're down to a very small team right now, so I'm going to defer this to another time.

PR #853, opened 2022-04-21
make setting the oauth client id for dev environments easier to find

rcpeters:
Make setting the oauth client id for dev environments easier to find by setting it to a unique string that is easy to search for.

### DIFF
--- a/settings/chrome-dev.json
+++ b/settings/chrome-dev.json
@@ -6,6 +6,8 @@
   "bouncerUrl": "http://localhost:8000/",
   "serviceUrl": "http://localhost:5000/",
 
+  "oauthClientId": "DEV_UNDEFINED_CLIENT",
+
   "browserIsChrome": true,
   "appType": "chrome-extension"
 }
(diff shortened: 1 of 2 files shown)


robertknight:
> Make setting the oauth client id for dev environments easier to find by setting it to a unique string that is easy to search for.

There are no changes to the documentation in the [instructions](https://github.com/hypothesis/browser-extension/blob/master/docs/building.md#building-for-a-local-h-server) for building the extension in this PR though. These currently say:

> Note: It may be convenient to create a new, untracked config file based on chrome-dev.json to maintain your settings. Any settings JSON file in the settings directory with custom in the name will be git-ignored (here called custom.json):

> Set an `oauthClientId` property in your settings JSON. Its value should be the 32-character ID of this newly-created OAuthClient.

When writing these instructions we had assumed that adding a property to a small JSON file would be an easy thing for readers to do.

robertknight:
Following up on my previous comment: I appreciate your interest in contributing, but this PR is currently incomplete and I don't see that it provides value in its current state, so I'm going to close it.

PR #907, opened 2022-06-20
Add GitHub Actions CI

robertknight:
This is a step towards moving builds away from Jenkins, which is flakey
when the build master is busy.


robertknight:
Discussed here - https://hypothes-is.slack.com/archives/C1M8NH76X/p1655730172300659

PR #975, opened 2022-09-21
Add script and GitHub Actions workflow to update client

robertknight:
The plan is for this workflow and script to be triggered at the end of a new Hypothesis client release, in order to automatically update the browser extension. This matches functionality that was previously implemented in the client/extension's Jenkins build pipelines.


robertknight:
I'm going to merge this and give it a trial run.

PR #979, opened 2022-09-22
Check out new tag when releasing updated client

robertknight:
When running the release workflow after updating the client, make sure that the new tag created by the update-client workflow is checked out, instead of the previous commit that was at the tip of the main branch when the update-client workflow was initiated.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,11 @@ concurrency:
 
 on:
   workflow_call:
+    inputs:
+      ref:
+        description: 'Git commit to release'
+        required: false
+        type: string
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -24,6 +29,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v2
+      with:
+        ref: ${{ inputs.ref }}
```
robertknight, line 33: I'm not sure if this will result in an error when the workflow is run directly or whether the default value will be used for this input, which is the behavior I want. Let's see...

PR #1031, opened 2022-11-14
Support building extension for Manifest V3

robertknight:
This PR enables the extension to be built as a Manifest V3 extension by setting the `manifestV3` key to `true` in the settings JSON.

**Summary of changes:**

- Add conditionals in `manifest.json` to build extension for Manifest V3
- Use MV3 APIs for executing scripts (files or functions) in the page when building for Manifest V3

**Testing:**

1. Add `"manifestV3": true` to the settings JSON file you use for development, and build the extension
2. Build the extension and load it in Chrome.
3. Go to the extension's info page under chrome://extension and under "Inspect views" you should see that it displays "Service worker" instead of background page

Example of my local Chrome settings JSON:

```json
{
  "buildType": "dev",
  "manifestV3": true,

  "apiUrl": "http://localhost:5000/api/",
  "authDomain": "localhost",
  "bouncerUrl": "http://localhost:8000/",
  "serviceUrl": "http://localhost:5000/",

  "browserIsChrome": true,
  "appType": "chrome-extension",

  "oauthClientId": "b1a421ec-fd9d-11ea-87b6-cb79e356a0c9"
}
```

### DIFF
--- a/src/background/chrome-api.js
+++ b/src/background/chrome-api.js
@@ -74,7 +74,7 @@ export function getChromeAPI(chrome = globalThis.chrome) {
    * @return {(...args: Args) => Promise<Result>}
    */
   const promisifyAlt = fn => {
-    // @ts-ignore
+    // @ts-expect-error
     return promisify(fn);
   };
 
@@ -85,10 +85,7 @@ export function getChromeAPI(chrome = globalThis.chrome) {
       onClicked: browserAction.onClicked,
       setBadgeBackgroundColor: promisify(browserAction.setBadgeBackgroundColor),
       setBadgeText: promisify(browserAction.setBadgeText),
-
-      // @ts-ignore - Ignore an incorrect typing error about setIcon's callback
       setIcon: promisify(browserAction.setIcon),
-
       setTitle: promisify(browserAction.setTitle),
     },
 
--- a/src/background/chrome-api.js
+++ b/src/background/chrome-api.js
@@ -242,11 +242,10 @@ export async function executeFunction(
     }
     const results = await chromeAPI_.scripting.executeScript({
       target,
-      // @ts-expect-error - Typechecking error needs debugging.
       func,
       args,
     });
-    return results[0].result;
+    return /** @type {Result} */ (results[0].result);
   }
 
   const code = codeStringForFunctionCall(func, args);
--- a/src/background/chrome-api.js
+++ b/src/background/chrome-api.js
@@ -8,6 +8,8 @@
  *  - Provide a seam that can be easily mocked in tests
  */
 
+import settings from './settings';
+
 /**
  * Wrap the browser APIs exposed via the `chrome` object to return promises.
  *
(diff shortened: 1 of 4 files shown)


robertknight:
It looks like we may need to dust off this PR in order to publish a new fork of the extension for the Confluence integration.

robertknight:
I found an issue with VitalSource in testing. Trying to activate the MV3 extension when viewing a VitalSource book at https://bookshelf.vitalsource.com leads to an error:

user-gesture-error

The error happens on this line:

https://github.com/hypothesis/browser-extension/blob/32d635505946442bb7834d9965afc0af328dbe1b/src/background/sidebar-injector.js#L434

The call to `chrome.permissions.request` here _is_ happening in response to a user gesture, but there are several async steps in-between. This worked in MV2 but it seems that user gestures are tracked differently with MV3.

robertknight:
I found a workaround for the VitalSource issue mentioned in the previous comment. Since the change to use Manifest V3 is behind a build-time feature flag that is off by default, that workaround can go in a separate PR.

src/background/chrome-api.js
```diff
@@ -72,19 +74,19 @@ export function getChromeAPI(chrome = globalThis.chrome) {
    * @return {(...args: Args) => Promise<Result>}
    */
   const promisifyAlt = fn => {
-    // @ts-ignore
+    // @ts-expect-error
     return promisify(fn);
   };
```
acelaya, line 79: Not related to this PR, but I just remembered I wanted to ask about this.

What's the purpose of `promisifyAlt`? Isn't it just forwarding to `promisify`? Would it be the same to just use `promisify`, or am I missing something?

robertknight: This is a typing hack that gives `promisify` a different signature. It avoids needing to mark every place where `promisifyAlt` is used as an expected error.

src/manifest.json.mustache
```diff
@@ -40,53 +46,100 @@
...
+  {{/manifestV3}}
+
+  {{^manifestV3}}
   "browser_action": {
     "default_icon": {
       "19": "images/browser-icon-inactive.png",
       "38": "images/browser-icon-inactive@2x.png"
     }
   },
+  {{/manifestV3}}
```
acelaya, line 115: Suggestion: maybe we could do this:

```
{{#manifestV3}}
"action": {
{{/manifestV3}}
{{^manifestV3}}
"browser_action": {
{{/manifestV3}}
  "default_icon": {
    "19": "images/browser-icon-inactive.png",
    "38": "images/browser-icon-inactive@2x.png"
  }
},
```

That way we define the icons only once and it's less likely to forget editing one.

On the other hand we probably get rid of most of this logic soon, so maybe it's not worth it.

acelaya (approved):
Looks good!

src/background/chrome-api.js
acelaya: ooohhh, I see, Isee. I didn't realize the jsdoc was different on them.

PR #1178, opened 2023-04-06
Remove .python-version file

robertknight:
The issue described in 5a4a4d5bed4b4f9236ca518f61ebf8c09f44b4cf appears to have been resolved, and it is no longer necessary to force the use of Python 2 with node-gyp.

To test:

1. `rm -rf node_modules`
2. `yarn install` should then complete without errors on a system where `python` is Python 3.


robertknight:
I'm going to go ahead and merge this since it is trivial and several of my colleagues are out.

PR #1188, opened 2023-04-10
Bump minimum Chrome version to 88

robertknight:
This is a prerequisite for publishing the extension as a Manifest V3 extension.

Chrome 88 was published in Jan 2021, so > 2 years old at this point. There are some MV3 APIs that require a newer version of Chrome than 88, but I haven't found a need for them yet.

Part of #622


acelaya (approved):
👍🏼

Issue #37, opened 2017-04-12
Extension fails to inject after enabling it, then activating on previously open tab


robertknight:
**Steps to reproduce:**

1. Install the extension but disable it from chrome://extensions by unchecking the "Enable" box next to the extension's details
2. Open a tab at some HTTP URL
3. Re-enable the extension
4. Switch back to the tab and click the extension's toolbar button

**Expected result:**

The extension activates and injects the client into the page

**Actual result:**

The extension's toolbar button activates but the client is not injected into the page. After reloading the page, the client is successfully injected.

The problem only affects tabs that were opened prior to re-enabling the extension.

seanh:
I was able to reproduce this effect (toolbar button activates, but no sidebar) today, but I was uninstalling and reinstalling the extension rather than enabling and disabling. It was only happening on tabs that were already open when I re-installed the extension, not happening on other tabs, and when I reloaded the broken tab the problem went away.

Also:

1. Quit Chrome
2. Re-launch Chrome, it opens all the tabs you had open last time
3. Try to launch h in one of those tabs => no sidebar. I seem to be able to reliably reproduce this on http://boingboing.net/ and some other sites, but reliably _don't_ reproduce it on https://lobste.rs/

Looks like Dan was also having a similar problem: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1497547469362566

No leads yet on what is actually happening

PR #1190, opened 2023-04-11
Replace `localStorage` with new approach to persisting tab state across extension reloads

robertknight:
Change how the Hypothesis-related state for each tab is initialized on
startup. Instead of caching the previous state in localStorage, instead
run code in each tab to query whether the client is currently loaded and
then mark the tab as active or inactive accordingly.

Fixes https://github.com/hypothesis/browser-extension/issues/37.

This change solves several problems:

 - It fixes an issue where the extension failed to inject into existing
   tabs after an installation, until the tab was navigated, because the
   `ready` state for these tabs was incorrectly initialized to `false`.
   This means that the extension incorrectly treated pre-existing tabs
   as being in a loading state.

 - It avoids the possibility of the saved state in the extension's
   localStorage getting out of sync with the actual state. This can
   happen if the tabs in the browser change at all while the extension's
   background page is not running.

 - It means that `TabStore`, and its use of localStorage, are no longer
   needed. localStorage is not available in Manifest V3 when the
   background page becomes a Service Worker, so we would have had to
   replace this with some async storage mechanism.

A downside to this approach is that code has to be run in each tab to determine
the client state. This would not work if the client lost permission to run code
in arbitrary tabs (eg. if browsers made it possible to revoke this required
permission in future). In that case it would have to assume that the client was
not active in the tab. Checking whether the client is active in a tab is done
with a simple CSS selector query, so we might be able to use some declarative
API as a replacement in that case.

As part of this change, the methods for initializing the `Extension`
class have been changed. `Extension.prototype.listen` has been replaced
with an async `init` method both installs listeners and queries the
state of existing tabs.

**TODO:**

- [x] Confirm this fixes https://github.com/hypothesis/browser-extension/issues/37
- [x] Add tests for `SidebarInjector.isClientActiveInTab`
- [x] Add tests for `Extension.init`


src/background/extension.js
```diff
@@ -94,42 +57,105 @@ export class Extension {
...
+        // nb. If tab status is not available, we optimistically assume it is
+        // loaded.
+        const ready =
+          tab.status === 'complete' || typeof tab.status !== 'string';
+
+        state.setState(tab.id, {
+          state: isActive ? 'active' : 'inactive',
+          extensionSidebarInstalled: isActive,
+          ready,
+        });
```
robertknight, line 102: Something to note here is that the only per-tab state we populate after reloading the extension is:

1. Whether the client is active
2. Whether the tab is loaded

Additional state such as cached annotation counts (on the badge) and injection errors (when the badge displays `!` instead of a count) are lost. I think that's OK. The errors will appear again if you retry activating the client and the badge will update once you interact with the tab again.

src/background/extension.js
```diff
@@ -6,67 +6,30 @@ import { HelpPage } from './help-page';
 import settings from './settings';
 import { SidebarInjector } from './sidebar-injector';
 import { TabState } from './tab-state';
-import { TabStore } from './tab-store';
 
 /**
```
robertknight, line 10: The previous comments here described many implementation details and consequently got out of date. The new comments just describe the overall purpose of the class and how to use it.

robertknight:
While looking through Chrome's "[What's new](https://developer.chrome.com/docs/extensions/whatsnew/)" page for extensions I saw a mention of a built-in API for testing whether an extension is [active in a tab](https://developer.chrome.com/docs/extensions/reference/action/#method-isEnabled). If this works it might save us from needing to implement anything ourselves.

src/background/extension.js
```diff
@@ -94,42 +57,105 @@ export class Extension {
...
+     * This queries the state of the extension in existing tabs and sets up
+     * event listeners to respond to future tab changes.
+     *
+     * If the extension's state in a particular tab cannot be determined,
+     * the extension is assumed not to be loaded in that tab.
+     *
+     * @return - A promise that resolves once listeners have been set up and
+     *   the state of existing tabs has been determined.
+     */
+    this.init = async () => {
```
acelaya, line 118: Maybe this could be defined as a regular class method instead of a callable property?

acelaya (approved):
Looks good.

src/background/extension.js
robertknight: In this PR I stuck to the existing style of this module, but in future this code would benefit from some modernization. You'll see this elsewhere in the browser-extension codebase too, places where pre-ES6-isms still exist.

Issue #1179, opened 2023-04-06
Client can be mis-configured due to config and boot-script injection being non-atomic


robertknight:
When testing changes to the integration between bouncer (hyp.is) and the extension locally, I encountered a situation where the development extension would sometimes be activated in the toolbar yet the client showed annotations from the production instance of h.

After adding some logging I realized the issue is that the extension injects the client's configuration (which tells it where to load resources from) in a separate step to starting the client, and that it is possible for the tab to navigate in-between these steps. When this happens the client can end up starting without the configuration being present on the page, causing it to load assets from the wrong location.

The sequence of events when this happens was:

1. Bouncer page A loads
2. Bouncer page A sends a message to the extension and asks it activate itself and navigate to URL B
3. The extension attempts to activate itself in the tab before the navigation has started (`SidebarInjector.injectIntoTab`)
4. The extension injects a configuration `<script>` tag into the page (start of `injectIntoRemoteDocument` function)
5. The tab navigates from A to B
6. The extension injects the client's boot script into the page (next step of `injectIntoRemoteDocument` function)
7. The extension gets a notification about the navigation from A to B having happened.
8. The extension re-injects the client configuration
9. The extension re-injects the client's boot script

<img width="692" alt="Client startup logging" src="https://user-images.githubusercontent.com/2458/230428294-e69e37c4-8acb-4ab4-91a4-5bfe0e5dcacc.png">

In this sequence, steps 4 and 6 are different steps of the same async function in `injectIntoRemoteDocument`, and are performed in the same browser tab, but they end up being executed in different documents since a navigation is happening. As a result when the client loads in step 6, it doesn't find its configuration and ends up loading with the default production configuration. Step 8 has no effect, since it happens after the client's boot script has already loaded. Step 9 has no effect, since the boot script will detect the presence of another Hypothesis client in the page and bail out.

Although this situation involved some local changes to the bouncer, I believe the same outcome is possible in normal usage if the page triggers a navigation at the same time as the extension is activated.

## Possible solutions

1. Somehow make steps 4 and 6 above a single atomic operation which are guaranteed to execute in the same document
2. Allow the re-injected client in step 9 to remove and replace the client injected in step 6
3. Make step 6 abort if the client configuration is not found in the page

robertknight:
[...]

PR #1193, opened 2023-04-11
Check hypothesis config script exists before trying to execute client boot script

acelaya:
This is an attempt on addressing #1179, by implementing the third of the suggested solutions there.

I first tried to make the config loading and boot script execution part of a single `chromeAPI.tabs.executeScript(...)` execution, but it required merging somehow code generated on the fly, with a JS script bundled with the extension, which is not supported out of the box.

I thought we could "read" the bundled file, then prepend the rest of the code, and then execute the whole resulting script as string, but there doesn't seem to be an API allowing to do that. Even if there was, it's quite messy.

Instead I have changed the behavior to make the boot script execution first check that the config script tag exists. This is now more reliable because we [remove that tag when the extension is deactivated](https://github.com/hypothesis/browser-extension/pull/1191).

This way, if there's a race condition that makes the config to be loaded first[^1], and then some other script running on the same tab triggers a client-side navigation to other page, and then we try to execute the boot script (which would happen in the new page, where the config script no longer exists), the last step will be aborted.

[^1]: This behavior corresponds to steps 4, 5 and 6 from #1179 description.

### DIFF
--- a/src/background/sidebar-injector.js
+++ b/src/background/sidebar-injector.js
@@ -27,6 +27,10 @@ function setClientConfig(config) {
   document.head.appendChild(script);
 }
 
+function hypothesisConfigExists() {
+  return !!document.querySelector('script.js-hypothesis-config[data-remove-on-unload]');
+}
+
 /**
  * A Chrome tab for which we have ID and URL information.
  *
@@ -262,10 +266,8 @@ export function SidebarInjector() {
         "Hypothesis extension can't be used on Hypothesis LMS assignments"
       );
     } else {
-      await injectConfig(tab.id, config);
-
       const result = /** @type {{ installedURL: string }|null} */ (
-        await injectIntoHTML(tab)
+        await injectIntoHTML(tab, config)
       );
       if (
         typeof result?.installedURL === 'string' &&
@@ -296,12 +298,13 @@ export function SidebarInjector() {
     }
   }
 
-  /** @param {Tab} tab */
-  function injectIntoHTML(tab) {
-    return executeScript({
-      tabId: tab.id,
-      file: '/client/build/boot.js',
-    });
+  /**
+   * @param {Tab} tab
+   * @param {object} config
+   */
+  async function injectIntoHTML(tab, config) {
+    await injectConfig(tab.id, config);
+    return executeClientBootScript(tab.id);
   }
 
   /** @param {Tab} tab */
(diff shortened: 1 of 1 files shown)


src/background/sidebar-injector.js
```diff
@@ -262,10 +266,8 @@ export function SidebarInjector() {
         "Hypothesis extension can't be used on Hypothesis LMS assignments"
       );
     } else {
-      await injectConfig(tab.id, config);
-
       const result = /** @type {{ installedURL: string }|null} */ (
-        await injectIntoHTML(tab)
+        await injectIntoHTML(tab, config)
```
acelaya, line 270: I have moved the call to `injectConfig` inside `injectIntoHTML`, to make it consistent with `injectIntoVitalSourceReader`.

src/background/sidebar-injector.js
```diff
@@ -27,6 +27,10 @@ function setClientConfig(config) {
   document.head.appendChild(script);
 }
 
+function hypothesisConfigExists() {
+  return !!document.querySelector('script.js-hypothesis-config[data-remove-on-unload]');
```
acelaya, line 31: I have tried to make the selector as specific as possible, but maybe `script.js-hypothesis-config` would be enough.

robertknight:
This potentially has a similar issue to before. Consider this sequence of events:

1. Config injected into tab displaying page A
2. Presence of config in tab is checked while still displaying page A. This returns true.
3. Tab navigates to page B
4. Boot script is injected into page B, where there is no config

If we can find a way for the boot script to identify where it is being loaded from (`chrome-extension://{id}/...`) then it could look for matching config, if loaded from certain URLs (eg. scheme is `moz-extension:` or `chrome-extension:`) and bail out if not found. This would implement the same idea, but without the race condition.

src/background/sidebar-injector.js
```diff
@@ -438,4 +437,39 @@ export function SidebarInjector() {
...
+   * script loads.
+   *
+   * @param {number} tabId
+   * @param {number} [frameId]
+   */
+  async function executeClientBootScript(tabId, frameId) {
+    const configFound = await configExists(tabId, frameId);
+    if (! configFound) {
+      return Promise.resolve();
+    }
```
robertknight, line 454: These two steps `await configExits(...)` and `executeScript(...)` are separate commands sent to the process where the page is displayed, so it is possible for navigations to happen in-between.

acelaya:
> This potentially has a similar issue to before.

You are absolutely right. I don't know how I came to the conclusion that this would do 😅 

I was trying to avoid changes in client's boot script itself, to make this completely handled by the browser extension, but I'll give it another round of thought.

PR #1194, opened 2023-04-12
Refactor sidebar-injector to remove duplication

acelaya:
Small refactor brought from https://github.com/hypothesis/browser-extension/pull/1193, which was not merged, but it introduced 2 changes that I think are worth keeping:

- Make `injectIntoHTML` implicitly call `injectConfig`, making it consistent with `injectIntoVitalSourceReader`.
- Wrap the execution of the client's boot script into its own reusable function.

### DIFF
--- a/src/background/sidebar-injector.js
+++ b/src/background/sidebar-injector.js
@@ -262,10 +262,8 @@ export function SidebarInjector() {
         "Hypothesis extension can't be used on Hypothesis LMS assignments"
       );
     } else {
-      await injectConfig(tab.id, config);
-
       const result = /** @type {{ installedURL: string }|null} */ (
-        await injectIntoHTML(tab)
+        await injectIntoHTML(tab, config)
       );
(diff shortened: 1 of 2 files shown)


src/background/sidebar-injector.js
```diff
@@ -438,4 +433,16 @@ export function SidebarInjector() {
...
   }
+
+  /**
+   * @param {number} tabId
+   * @param {number} [frameId]
+   */
+  async function executeClientBootScript(tabId, frameId) {
+    return executeScript({
+      tabId,
+      frameId,
```
robertknight, line 444: I assume Chrome ignores the `frameId` if the property is set but `undefined`?

robertknight (approved):

src/background/sidebar-injector.js
acelaya: Yes

PR #1195, opened 2023-04-12
Request `webNavigation` permission directly after clicking browser action

robertknight:
Work around an issue in Manifest V3 where async calls in-between a user clicking an extension's toolbar action and attempting to request additional permissions results in an error "The function must be called during a user gesture" when calling `chrome.permissions.request` [1]. In our context this happens when activating Hypothesis on a https://bookshelf.vitalsource.com URL.

The workaround is to request any additional permissions that might be required, based on the tab's URL, immediately after clicking the toolbar action. This workaround relies on us being able to determine what permissions may be needed based on the URL alone, which is currently the case.

In an additional commit I also fixed an issue where error messages of > 60 chars would overflow their container on the extension error info screen.

[1] https://bugs.chromium.org/p/chromium/issues/detail?id=1363490
[2] https://github.com/hypothesis/browser-extension/pull/1031#issuecomment-1504955470

**Testing:**

1. Build a Manifest V3 version of the extension following steps in https://github.com/hypothesis/browser-extension/pull/1031
2. Go to https://bookshelf.vitalsource.com and login (there are credentials in the usual place if you don't have an account)
3. Open a book
4. Activate the extension

After granting permissions, you can test the scenario where the permissions are rejected by removing and re-adding the extension. There might be an easier way to reset granted permissions, but I haven't found it yet.

On the branch in https://github.com/hypothesis/browser-extension/pull/1031 you'll get [an error](https://github.com/hypothesis/browser-extension/pull/1031#issuecomment-1504955470) when attempting to activate the client. On this branch you should get a permissions prompt and then, if you accept that, the client should inject.

### DIFF
--- a/src/background/extension.js
+++ b/src/background/extension.js
@@ -162,7 +162,7 @@ export class Extension {
     this._onTabStateChange = onTabStateChange;
 
     /** @param {chrome.tabs.Tab} tab */
-    function onBrowserActionClicked(tab) {
+    async function onBrowserActionClicked(tab) {
       const tabId = /** @type {number} */ (tab.id);
       const tabError = state.getState(tabId).error;
       if (tabError) {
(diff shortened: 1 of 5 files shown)


acelaya (approved):
👍🏼

Issue #1177, opened 2023-04-05
Only activate appropriate version of extension when following a bouncer link


robertknight:
The [Bouncer](https://github.com/hypothesis/bouncer) service handles annotation sharing links (`https://hyp.is/...`)

1. Checking for the presence of a specific Chrome extension ID
2. If found, redirecting the user to the annotation's original URL with `#annotations:{annotation_id}` appended

Step (1) allows differentiating between different flavors / forks of the Hypothesis extension depending on the annotation, but step 2 does not.

This is a problem when developing Hypothesis because bouncer links from dev, QA or production versions of bouncer will try to activate _all_ of the Hypothesis extensions installed in step 2. One of these will non-deterministically win, which might be the wrong extension for the bouncer service/annotation.

This is also a problem for a new variant of the extension we are building which integrates Hypothesis + Confluence. The bouncer service we use for that, whether the existing one or a new service, has no way of only triggering the Confluence-integrated extension.

## Possible solutions

1. Change the fragment identifier that triggers the extension, so that it indicates the extension ID. eg. `#annotations:(annotation_id}@{extension_id}`
2. Change the way the bouncer service works so that instead of redirecting to a URL, it communicates with the specific target extension, and the extension in turn navigates the current tab and activates the client afterwards

robertknight:
This issue can be closed once https://github.com/hypothesis/browser-extension/pull/1208 is deployed.

PR #1197, opened 2023-04-13
Enable messages from bouncer to activate the extension

robertknight:
Add support for a new external message, "activate", which enables the extension on the tab that sent the message, optionally redirecting the tab to a new URL first. The existing "ping" message also gains support for a `queryFeatures` field which allows bouncer to test whether the installed extension supports the "activate" message.

This enables the bouncer service to activate a specific Hypothesis extension directly, as opposed to having all installed Hypothesis extensions watch and respond to the presence of an `#annotations:` fragment in the URL. The latter approach causes a conflict when multiple Hypothesis extensions are installed.

Part of #1177


src/background/extension.js
```diff
@@ -103,39 +175,6 @@ export class Extension {
...
-     * This queries the state of the extension in existing tabs and sets up
-     * event listeners to respond to future tab changes.
-     *
-     * If the extension's state in a particular tab cannot be determined,
-     * the extension is assumed not to be loaded in that tab.
-     *
-     * @return - A promise that resolves once listeners have been set up and
-     *   the state of existing tabs has been determined.
-     */
-    this.init = async () => {
```
robertknight, line 118: This function was moved down the file to avoid a reference-before-declaration error after changing some `function` expressions to arrow callbacks, to avoid confusion with `this` references.

src/background/install.ts
```diff
@@ -0,0 +1,66 @@
...
+  // Respond to messages sent by the JavaScript from https://hyp.is.
+  // This is how it knows whether the user has this Chrome extension installed.
+  chromeAPI.runtime.onMessageExternal.addListener(
+    (request: ExternalMessage, sender, sendResponse) => {
+      switch (request.type) {
+        case 'ping':
+          {
+            const queryFeatures = request.queryFeatures ?? [];
+            const features = allFeatures.filter(f => queryFeatures.includes(f));
+            sendResponse({ type: 'pong', features });
```
robertknight, line 34: The `queryFeatures` request property and `features` response property is new here.

src/background/install.ts
```diff
@@ -0,0 +1,66 @@
...
+            const { url, query } = request;
+            if (url) {
+              chromeAPI.tabs.update(sender.tab.id, { url });
+            }
+            extension.activate(sender.tab.id, {
+              afterNavigationTo: url,
+              query,
+            });
+
+            sendResponse({ active: true });
```
robertknight, line 52: The `{ active: true }` response here is not actually used by bouncer, it just serves to confirm the request was handled.

acelaya (approved):
👍🏼

PR #1204, opened 2023-04-17
Convert several modules in src/background/ to TypeScript

robertknight:
Convert the following modules to TS:

```
src/background/
  browser-action.js
  direct-link-query.js
  extension.js
  index.js
  help-page.js
  settings.js
```

The three main remaining modules are chrome-api.js, tab-state.js and sidebar-injector.js. These require more work, so will be done separately.


src/background/settings.ts
```diff
@@ -3,24 +3,24 @@
...
  * Configuration data for the extension.
  */
-export default /** @type {Settings} */ ({
+export default {
   ...settings,
 
   // Ensure API url does not end with '/'
   apiUrl: settings.apiUrl.replace(/\/$/, ''),
-});
+} as Settings;
```
acelaya, line 26: Perhaps we could use `satisfies Settings` instead? Otherwise it will not complain even if the object does not actually fulfill `Settings`, which could lead to unintentional errors later on, due to typos and such.

```suggestion
} satisfies Settings;
```

acelaya (approved):
Looks good. Just suggested a small change.

src/background/settings.ts
robertknight: >  Otherwise it will not complain even if the object does not actually fulfill Settings

True. I modified this file so that the exported type is exactly `Settings`, but the conversion is done in a way that catches errors (`const settings: Settings = { ... }; export default settings;`).

PR #1206, opened 2023-04-17
Convert SidebarInjector to TypeScript

robertknight:
Convert sidebar-injector.js to TS and do a minimal class-ification in the process. There is further work required to make it a conventional ES class (ie. without all the code in the constructor).

While working on this I realized we really ought to rename this as well, because it injects the whole Hypothesis client into a tab, not just the sidebar container. This can happen later.


src/background/sidebar-injector.ts
```diff
@@ -0,0 +1,485 @@
...
+  BlockedSiteError,
+  LocalFileError,
+  NoFileAccessError,
+  RestrictedProtocolError,
+} from './errors';
+
+const CONTENT_TYPE_HTML = 'HTML';
+const CONTENT_TYPE_PDF = 'PDF';
+const CONTENT_TYPE_VITALSOURCE = 'VITALSOURCE';
+const CONTENT_TYPE_LMS = 'LMS';
```
acelaya, line 14: Maybe these could be "merged" into an `enum`.

src/background/sidebar-injector.ts
```diff
@@ -0,0 +1,485 @@
...
+ *
+ * All "normal" tabs should have this information because of the extension's
+ * permissions.
+ */
+function checkTab(tab: chrome.tabs.Tab): Tab {
+  if (!tab.id || !tab.url) {
+    throw new Error('Tab is missing ID or URL');
+  }
+  return tab as Tab;
+}
```
acelaya, line 59: Perhaps we could use a TypeScript assertion here (like a guard, but it throws instead of returning `false`).

```suggestion
function assertTab(tab: chrome.tabs.Tab): asserts tab is Tab {
  if (!tab.id || !tab.url) {
    throw new Error('Tab is missing ID or URL');
  }
}

function checkTab(tab: chrome.tabs.Tab): Tab {
  assertTab(tab);
  return tab;
}
```

This allows to remove the `as Tab` casting, as the assertion tells TS it has the proper type after that check.

But it's more verbose, so I'll leave it up to you.

src/background/sidebar-injector.ts
```diff
@@ -0,0 +1,485 @@
...
+
+    function getPDFViewerURL(url: string) {
+      // Encode the original URL but preserve the fragment, so that a
+      // '#annotations' fragment in the original URL will persist and trigger the
+      // sidebar to focus and scroll to that annotation when the PDF viewer loads.
+      const parsedURL = new URL(url);
+      const hash = parsedURL.hash;
+      parsedURL.hash = '';
+      const encodedURL = encodeURIComponent(parsedURL.href);
+      return pdfViewerBaseURL + '?file=' + encodedURL + hash;
```
acelaya, line 184:
```suggestion
      return `${pdfViewerBaseURL}?file=${encodedURL}${hash}`;
```

src/background/sidebar-injector.ts
```diff
@@ -0,0 +1,485 @@
...
+     * a tab with a given URL.
+     */
+    async function canInjectScript(url: string) {
+      if (isSupportedURL(url)) {
+        return true;
+      } else if (isFileURL(url)) {
+        return chromeAPI.extension.isAllowedFileSchemeAccess();
+      } else {
+        return false;
+      }
```
acelaya, line 198: I'm not completely sure if this is equivalent, but I think this could be:

```suggestion
      if (isFileURL(url)) {
        return chromeAPI.extension.isAllowedFileSchemeAccess();
      }

      return isSupportedURL(url);
```

But this is not related with the TS migration itself.

src/background/sidebar-injector.ts
```diff
@@ -0,0 +1,485 @@
...
+        // Chrome does not permit extensions to inject content scripts
+        // into (chrome*):// URLs and other custom schemes.
+        //
+        // A common case where this happens is when the user has an
+        // extension installed that provides a custom viewer for PDFs
+        // (or some other format). In some cases we could extract the original
+        // URL and open that in the Hypothesis viewer instead.
+        const protocol = tab.url.split(':')[0];
+        throw new RestrictedProtocolError(
+          'Cannot load Hypothesis into ' + protocol + ' pages'
```
acelaya, line 305:
```suggestion
          `Cannot load Hypothesis into ${protocol} pages`
```

src/background/sidebar-injector.ts
```diff
@@ -0,0 +1,485 @@
...
+        // The extension is blocked on LMS assignments to avoid confusion with the
+        // embedded Hypothesis instance. The user can still use the extension on other
+        // pages hosted in the LMS itself.
+        throw new BlockedSiteError(
+          "Hypothesis extension can't be used on Hypothesis LMS assignments"
+        );
+      } else {
+        const result = (await injectIntoHTML(tab, config)) as {
+          installedURL: string;
+        } | null;
```
acelaya, line 325: Probably easier to read if the return type is defined in the `injectIntoHTML` function, and in case we need to do some type casting, we do it there, instead of here.

src/background/sidebar-injector.ts
```diff
@@ -0,0 +1,485 @@
...
+      if (type === CONTENT_TYPE_PDF) {
+        await injectIntoPDF(tab);
+      } else if (type === CONTENT_TYPE_VITALSOURCE) {
+        await injectIntoVitalSourceReader(tab, config);
+      } else if (type === CONTENT_TYPE_LMS) {
+        // The extension is blocked on LMS assignments to avoid confusion with the
+        // embedded Hypothesis instance. The user can still use the extension on other
+        // pages hosted in the LMS itself.
+        throw new BlockedSiteError(
+          "Hypothesis extension can't be used on Hypothesis LMS assignments"
```
acelaya, line 320:
```suggestion
          'Hypothesis extension can't be used on Hypothesis LMS assignments'
```

acelaya (approved):
Added some comments and suggestions, but nothing critical. Feel free to address them separately, or just ignore them.

src/background/sidebar-injector.ts
robertknight: The text contains a single quote. That is why prettier formatted it this way.

src/background/sidebar-injector.ts
robertknight: The suggestion makes sense. I looked into this though and realized there was a bigger problem - nothing is actually setting `installedURL`! It looks like it used to be part of a copy of the client's boot script that was embedded into this repository - e0bf3fd2a09414170eb991d7837bf6acd821502b.

src/background/sidebar-injector.ts
acelaya: Oh, right. I didn't see that. In fact, the code highlight breaks on my suggestion 😅

src/background/sidebar-injector.ts
robertknight: Yes. I think I'd like to leave that as a separate task since it involves changes across several modules.

PR #1208, opened 2023-04-18
Pass same client configuration to PDFs as for other document types

robertknight:
Previously the Hypothesis client config passed to
`SidebarInjector.injectIntoTab` was not used when activating the client in a
PDF. Instead the client used static configuration from pdfjs-init.js. This
caused several problems:

 - When modifying the client config in extension.ts, the developer had to
   remember to update it in pdfjs-init.js separately (see
   96c1dca03492c8e44573d4366a449050d898b3cd for an example of this being
   forgotten).

 - There was no nice mechanism available to pass dynamic configuration to the
   client for PDF documents.

 - Passing of direct link query information from the extension to the client
   worked differently in PDFs than HTML documents. This resulted in direct links
   for PDFs breaking after 3df9fc0ba8f4f1c02f2a004a86390d81c458c282.

This commit resolves the issue by using extension messaging APIs to propagate
the client config passed to `SidebarInjector.injectIntoTab` into the PDF.js
init code in pdfjs-init.js.

Part of https://github.com/hypothesis/browser-extension/issues/1177.

**Testing:**

1. Make sure that annotating PDFs using the extension works
2. Test that direct links to annotations on PDFs works (make sure you have the latest version of the `bouncer` service checked out, and have set the `CHROME_EXTENSION_ID` env var when running it)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ extension: build/unload-client.js
 extension: build/pdfjs-init.js
 extension: $(addprefix build/,$(EXTENSION_SRC))
 
-build/extension.bundle.js: src/background/*.js rollup.config.mjs build/settings.json
+build/extension.bundle.js: src/background/*.js src/background/*.ts rollup.config.mjs build/settings.json
 	$(ROLLUP) -c rollup.config.mjs
 build/manifest.json: src/manifest.json.mustache build/settings.json
 	$(MUSTACHE) build/settings.json $< > $@
--- a/src/pdfjs-init.js
+++ b/src/pdfjs-init.js
@@ -5,35 +5,38 @@
 // This script is run once PDF.js has loaded and it configures the viewer
 // and injects the Hypothesis client.
 
-// Configure Hypothesis client to load assets from the extension instead of
-// the CDN.
-//
-// Note this configuration is duplicated in `src/background/extension.js`. Any
-// changes made here must be reflected there as well.
-const clientConfig = {
-  assetRoot: '/client/',
-  sidebarAppUrl: '/client/app.html',
-  notebookAppUrl: '/client/notebook.html',
-  profileAppUrl: '/client/profile.html',
-};
-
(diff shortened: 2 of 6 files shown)


src/pdfjs-init.js
```diff
@@ -5,35 +5,38 @@
...
+  // to finish initializing.
+  const [config] = await Promise.all([configPromise, viewerLoaded]);
+
+  const configScript = document.createElement('script');
+  configScript.type = 'application/json';
+  configScript.className = 'js-hypothesis-config';
+  configScript.textContent = JSON.stringify(config);
+
+  // This ensures the client removes the script when the extension is deactivated
+  configScript.setAttribute('data-remove-on-unload', '');
```
robertknight, line 34: I copied this from the existing code, although it is strictly redundant in this context since de-activating the extension on a PDF will navigate away from our PDF viewer to the original PDF URL.

Makefile
```diff
@@ -57,7 +57,7 @@ extension: build/unload-client.js
 extension: build/pdfjs-init.js
 extension: $(addprefix build/,$(EXTENSION_SRC))
 
-build/extension.bundle.js: src/background/*.js rollup.config.mjs build/settings.json
+build/extension.bundle.js: src/background/*.js src/background/*.ts rollup.config.mjs build/settings.json
```
acelaya, line 60: If this is a glob pattern, maybe we could "extend" the existing path to support both extensions.

```suggestion
build/extension.bundle.js: src/background/*.{js,ts} rollup.config.mjs build/settings.json
```

acelaya (approved):
👍🏼

Makefile
robertknight: This glob syntax isn't supported by Make. I'm not aware of a direct equivalent, though `src/background/*` probably would work fine here.

acelaya: > This glob syntax isn't supported by Make

That's too bad! 😅

PR #1211, opened 2023-04-19
Fix TS issues

acelaya:
Add `ts` files to tsconfig, and fix a couple type issues (most of them via type casting, `any` or `unknown`, so they need to be revisited).

### DIFF
--- a/src/background/chrome-api.ts
+++ b/src/background/chrome-api.ts
@@ -200,7 +200,10 @@ export async function executeScript(
     return results[0].result;
   }
 
-  const result = await chromeAPI_.tabs.executeScript(tabId, { frameId, file });
+  const result = (await chromeAPI_.tabs.executeScript(tabId, {
+    frameId,
+    file,
+  })) as unknown[];
   return result[0];
 }
 
@@ -239,7 +242,10 @@ export async function executeFunction<Args extends unknown[], Result>(
   }
 
   const code = codeStringForFunctionCall(func, args);
-  const result = await chromeAPI_.tabs.executeScript(tabId, { frameId, code });
+  const result = (await chromeAPI_.tabs.executeScript(tabId, {
+    frameId,
+    code,
+  })) as Result[];
   return result[0];
 }
 
--- a/src/background/detect-content-type.ts
+++ b/src/background/detect-content-type.ts
@@ -1,7 +1,9 @@
+export type PDFContentType = { type: 'PDF' };
+
+export type HTMLContentType = { type: 'HTML' };
+
 /** Details of the detected content type. */
-export type ContentTypeInfo = {
-  type: 'HTML' | 'PDF';
-};
+export type ContentTypeInfo = PDFContentType | HTMLContentType;
 
 /**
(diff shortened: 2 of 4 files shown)


src/background/chrome-api.ts
```diff
@@ -200,7 +200,10 @@ export async function executeScript(
     return results[0].result;
   }
 
-  const result = await chromeAPI_.tabs.executeScript(tabId, { frameId, file });
+  const result = (await chromeAPI_.tabs.executeScript(tabId, {
+    frameId,
+    file,
+  })) as unknown[];
```
acelaya, line 206: Nor putting a lot of effort into this, as this is part of Manifest v2 and getting removed soon.

src/background/chrome-api.ts
```diff
@@ -239,7 +242,10 @@ export async function executeFunction<Args extends unknown[], Result>(
   }
 
   const code = codeStringForFunctionCall(func, args);
-  const result = await chromeAPI_.tabs.executeScript(tabId, { frameId, code });
+  const result = (await chromeAPI_.tabs.executeScript(tabId, {
+    frameId,
+    code,
+  })) as Result[];
```
acelaya, line 248: Nor putting a lot of effort into this, as this is part of Manifest v2 and getting removed soon.

src/background/detect-content-type.ts
```diff
@@ -1,7 +1,9 @@
+type PDFContentType = { type: 'PDF' };
+
+type HTMLContentType = { type: 'HTML' };
```
robertknight, line 3: Since these used as part of an exported type then I think it makes sense to export them as well.

src/background/sidebar-injector.ts
```diff
@@ -283,10 +283,10 @@ export class SidebarInjector {
       return ['http:', 'https:', 'ftp:'].includes(parsedURL.protocol);
     }
 
-    async function injectIntoLocalDocument(tab: Tab) {
+    async function injectIntoLocalDocument(tab: Tab, config: object) {
       const type = await detectTabContentType(tab);
       if (type === CONTENT_TYPE_PDF) {
-        return injectIntoLocalPDF(tab);
+        return injectIntoLocalPDF(tab, config);
```
robertknight, line 289: Ah... good to see this spotted.

src/background/sidebar-injector.ts
```diff
@@ -347,7 +347,7 @@ export class SidebarInjector {
       }
 
       const onMessage = chromeAPI.runtime.onMessage;
-      const listener = (request, sender, sendResponse) => {
+      const listener = (request: any, sender: any, sendResponse: any) => {
```
robertknight, line 350: The Chrome extensions types that we use might already have a more specific type that is appropriate for `listener` here. If not, this will be OK.

robertknight (approved):

src/background/sidebar-injector.ts
acelaya: Good point. I will check that.

PR #1212, opened 2023-04-19
Set data-extension-id in config, when loading for PDFs

acelaya:


robertknight:
Looks good to me. I think `chrome.runtime.id` is available in this context though.

acelaya:
> Looks good to me. I think `chrome.runtime.id` is available in this context though.

It is. I have changed it to use `chrome.runtime.id` instead.

PR #1214, opened 2023-04-20
Modernize code using async/await, object shorthand syntax and arrow functions

robertknight:
This PR contains two general code style modernization updates.

- The first commit refactors the function that loads/unloads the client into tabs to use async / await, and also gives it a more obvious name
- The second commit enables the ESLint functions to prefer arrow functions over un-named `function`s (named functions are still allowed) and prefer object shorthand syntax over `foo: foo`. These lint rules are already enabled in client code.

### DIFF
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -72,7 +72,7 @@ export class Extension {
     const help = new HelpPage();
     const state = new TabState(onTabStateChange);
     const browserAction = new BrowserAction();
-    const sidebar = new SidebarInjector();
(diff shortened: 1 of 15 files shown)


src/background/extension.ts
```diff
@@ -183,7 +183,7 @@ export class Extension {
 
       browserAction.update(tabId, current);
 
-      updateTabDocument(tab);
+      addOrRemoveClientFromTab(tab);
```
acelaya, line 186: Nice 😅

acelaya (approved):
👏🏼

PR #1216, opened 2023-04-20
Remove raven-js crash reporting from Service Worker

robertknight:
Remove crash reporting infrastructure from the Service Worker. It was based on the old raven-js client library for Sentry and was not actually used. To make it work we would need to migrate to the new Sentry client libraries and do a bunch of other work.  Since it is not currently functional we can remove it in the interim and avoid maintaining it.


Makefile
```diff
@@ -57,7 +57,7 @@ extension: build/unload-client.js
 extension: build/pdfjs-init.js
 extension: $(addprefix build/,$(EXTENSION_SRC))
 
-build/extension.bundle.js: src/background/*.js src/background/*.ts rollup.config.mjs build/settings.json
+build/extension.bundle.js: src/background/*.ts rollup.config.mjs build/settings.json
```
robertknight, line 60: There are no longer any .js files in src/background, so remove the wildcard to avoid an error from make.

acelaya (approved):

PR #1217, opened 2023-04-20
Refactor `SidebarInjector` to a more conventional class and rename to `ClientInjector`

robertknight:
This class handles loading and unloading the whole Hypothesis client into a tab. The sidebar is just one component of the client, so `ClientInjector` is a more appropriate/accurate name. The class documentation has also been updated accordingly.

Internally this class had all of its logic in the constructor, including method definitions via property assignments. This comes from its pre-ES6 heritage. Refactor it to a more conventional class, which should be easier to navigate around than one very large function. It also allows tooling to understand it better.

The tests are unchanged except for replacing references to "sidebar" with "client".

### DIFF
--- a/src/background/client-injector.ts
+++ b/src/background/client-injector.ts
@@ -124,11 +124,14 @@ async function canInjectScript(url: string) {
 }
 
 /**
- * The SidebarInjector is used to deploy and remove the Hypothesis sidebar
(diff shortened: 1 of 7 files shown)


robertknight:
I'm going to close this for now while I'm working on the video player project and revisit this another time.

PR #1327, opened 2023-07-31
Add an option to auto toggle extension on certain websites.

aur3l14no:
This PR changes the options page and add two options: activated by default and Excluded URLs.

The extension is automatically activated on a tab when

- "activated by default" is on and the tab's URL does not match any excluded URLs
- "activated by default" is off and the tab's URL does match some excluded URL

image


robertknight:
Thanks for the proposal. We're a little short on bandwidth at the moment to work on this. In the interim, the best available alternative is to assign a keyboard shortcut to activate the extension.

I'll give some general thoughts on the UI. I think changing the meaning of "Excluded URLs" depending on the "Activate extension by default" checkbox is likely to prove confusing, evidenced by the fact that a paragraph of explanation was needed here. Also you'd have to rewrite the list if you decide to change the default. I would also be inclined to use a simpler pattern syntax for URLs by default that is familiar to more users (eg. just wildcards).

Concretely, I'm thinking of something more like this:

```
- Activate Hypothesis
 - ( ) When I click the extension icon
 - ( ) Automatically in every supported tab

Always activate automatically on these URLs:
[ < List of URL patterns > ]

Never activate automatically on these URLs:
[ < List of URL patterns > ]
```

Alternately the configuration could be expressed with a default on/off setting plus a set of custom rules which each specify a URL pattern and a behavior. That would require a more complex UI though.

Issue #1404, opened 2023-11-09
readme doc need update the newest command of `yarn link`


Mowmowj:
The README doc's command `yarn link` not work in client repo anymore. plz update it. thanks

PR #1405, opened 2023-11-09
Update `yarn link` instructions for Yarn v3+

robertknight:
The use of [`yarn link`](https://yarnpkg.com/cli/link) changed after the upgrade to Yarn v3+. It is no longer necessary to run `yarn link` in the client repository. Instead you can just do `yarn link ../path/to/client` in the browser extension repository.

`yarn unlink` still accepts a package name, so `yarn unlink hypothesis` still works. However the docs mention that you can also specify the path instead: https://yarnpkg.com/cli/unlink (ie. `yarn unlink ../client`).

Fixes https://github.com/hypothesis/browser-extension/issues/1404

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,20 +39,16 @@ The extension code has a test suite, which you can run using:
 
 Note that the browser extensions are for the most part just a wrapper around the
 [Hypothesis client][client]. Depending on what you're interested in working on,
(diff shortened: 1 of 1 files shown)


acelaya (approved):
LGTM

PR #1452, opened 2024-01-26
Update configuration for QA/staging extension

robertknight:
Update the browser extension following the replacement of qa.hypothes.is with staging.hypothes.is.

New auth clients have been registered at https://staging.hypothes.is/admin/oauthclients.

**Testing:**

1. Remove Hypothesis QA/staging extension if already installed
2. Build local staging extension:
   ```
   make build SETTINGS_FILE=settings/chrome-staging.json
   ```
3. Load extension in Chrome and verify that you can log in and annotate

The extension now displays the letters "STG" rather than "QA" on the badge.

### DIFF
--- a/src/background/sidebar-injector.ts
+++ b/src/background/sidebar-injector.ts
@@ -223,7 +223,7 @@ export class SidebarInjector {
 
     function isLMSAssignmentURL(url: string) {
       const { origin } = new URL(url);
-      // Matches origins like `lms.hypothes.is`, `qa-lms.hypothes.is`, `lms.ca.hypothes.is`.
+      // Matches origins like `lms.hypothes.is`, `staging-lms.hypothes.is`, `lms.ca.hypothes.is`.
       return /\blms\b/.test(origin) && origin.endsWith('.hypothes.is');
     }
 
--- a/tests/background/sidebar-injector-test.js
+++ b/tests/background/sidebar-injector-test.js
@@ -436,7 +436,7 @@ describe('SidebarInjector', () => {
         try {
           await injector.injectIntoTab({
             id: 1,
-            url: 'https://qa-lms.ca.hypothes.is/lti_launches',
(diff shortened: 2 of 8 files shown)


src/background/sidebar-injector.ts
```diff
@@ -223,7 +223,7 @@ export class SidebarInjector {
 
     function isLMSAssignmentURL(url: string) {
       const { origin } = new URL(url);
-      // Matches origins like `lms.hypothes.is`, `qa-lms.hypothes.is`, `lms.ca.hypothes.is`.
+      // Matches origins like `lms.hypothes.is`, `staging-lms.hypothes.is`, `lms.ca.hypothes.is`.
```
acelaya, line 226: I think the staging URL for the LMS is `lms.staging.hypothes.is`

The logic would still work.

tests/background/sidebar-injector-test.js
```diff
@@ -436,7 +436,7 @@ describe('SidebarInjector', () => {
         try {
           await injector.injectIntoTab({
             id: 1,
-            url: 'https://qa-lms.ca.hypothes.is/lti_launches',
+            url: 'https://staging-lms.ca.hypothes.is/lti_launches',
```
acelaya, line 439: As per my previous comment, I think this URL might not be correct, but I can't find what would be the right one, if `lms.staging.ca.hypothes.is` or `lms.ca.staging.hypothes.is`

acelaya (commented):
I noticed a couple URLs that may not be correct. Let's double check those.

src/background/sidebar-injector.ts
robertknight: That's actually fine. It is intentional that these are only example origins and might differ if we decide to change the domain name syntax slightly for whatever reason, as long as we still include "lms" somewhere.

acelaya (approved):

PR #1507, opened 2024-04-15
Convert Node scripts to module scripts

robertknight:
This reduces the need for variation in eg. linting rules by making all JS files modules where possible. Note that Karma does not support this. See https://github.com/karma-runner/karma/issues/3677.


tools/settings.js
```diff
@@ -5,10 +5,11 @@
  * file and the package environment.
  */
 
-'use strict';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
```
robertknight, line 9: While writing this I realized we didn't have a convention about whether to use the `node:` prefix or not and whether it was considered preferred practice. https://github.com/nodejs/node/issues/38343 suggests that using the prefix for built-in modules is preferred.

PR #1588, opened 2024-09-05
Support OOPIF PDF viewer in Chrome

robertknight:
Fix detection of document type in Chrome's built-in PDF viewer when the "OOPIF
for PDF Viewer" feature ("pdf-oopif") is enabled.

Fixes https://github.com/hypothesis/support/issues/145

Related upstream issue tracking launch of this Chrome change: https://issuetracker.google.com/issues/40268279

----

**Testing:**

1. In Chrome, go to chrome://flags and enable the "OOPIF PDF viewer" experiment.
2. Restart your browser
3. Build extension on main, navigate to https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf and attempt to activate the extension. It should fail to activate.
4. Check out this branch and repeat step (3). The PDF viewer should work as normal.
5. Disable the "OOPIF PDF viewer" experiment in Chrome, restart the browser and very that this branch still works with the old PDF viewer.

### DIFF
--- a/src/background/chrome-api.ts
+++ b/src/background/chrome-api.ts
@@ -16,8 +16,9 @@ type Callback<Result> = (r: Result) => void;
  * This is exposed for testing. Consumers should use {@link chromeAPI}.
  */
 export function getChromeAPI(chrome = globalThis.chrome) {
-  // In the test environment, the `chrome` global may not exist.
-  if (typeof chrome === 'undefined') {
+  // In the test environment, the `chrome` global may not exist or may not
+  // be the expected object.
+  if (typeof chrome === 'undefined' || !chrome.extension) {
     // The `as never` causes this branch to be ignored when TS determines the
     // return type of this function.
     return null as never;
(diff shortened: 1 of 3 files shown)


src/background/chrome-api.ts
```diff
@@ -16,8 +16,9 @@ type Callback<Result> = (r: Result) => void;
  * This is exposed for testing. Consumers should use {@link chromeAPI}.
  */
 export function getChromeAPI(chrome = globalThis.chrome) {
-  // In the test environment, the `chrome` global may not exist.
-  if (typeof chrome === 'undefined') {
+  // In the test environment, the `chrome` global may not exist or may not
+  // be the expected object.
+  if (typeof chrome === 'undefined' || !chrome.extension) {
```
robertknight, line 21: This change fixes test failures I encountered on main when the `chromeAPI` variable in this file is initialized. Previously the `chrome` global did not exist at all in the ordinary web pages where the test runs. In Chrome Dev this global _is_ defined in that environment, but is missing most of the APIs that are expected in the extension environment.

acelaya (approved):
LGTM

PR #1607, opened 2024-10-02
Simplify Chrome API wrapper

robertknight:
Since the migration to Manifest V3 is now complete, all of the `chrome` APIs support returning promises, so we don't need to wrap the older callback-returning APIs to return promises. This simplifies the code and also removes the need for a TypeScript workaround when resolving the type of wrapped functions.

In the process I had to fix an error when running `make format` locally due to `node_modules/.bin/prettier` not being executable. This had been on my TODO list for a while.


acelaya (approved):
👍🏼

PR #1617, opened 2024-10-17
Migrate to ESLint 9 and ESLint flat config

acelaya:
Update project to ESLint 9 and migrate ESlint config to flat format.

For this migration some eslint plugins needed to be updated:
    * `@typescript-eslint/...` packages replaced by `typescript-eslint` and updated to v8
    * `eslint-plugin-react-hooks` updated to v5
    * `eslint-plugin-mocha` updated to v10.5
    * `eslint-plugin-react` updated to v7.37
    * `eslint-config-hypothesis` updated to v3


eslint.config.js
```diff
@@ -0,0 +1,83 @@
...
+    ignores: [
+      '.yalc/**/*',
+      '.yarn/**/*',
+      'build/**/*',
+      'dist/**/*',
+      '**/vendor/**/*.js',
+      '**/coverage/**/*',
+      'docs/_build/*',
+      // TODO - Lint these files
+      'rollup*.config.js',
```
acelaya, line 16: I kept these ignored as it was before. We can change this later.

package.json
```diff
@@ -19,19 +19,18 @@
...
-    "eslint": "^8.57.0",
-    "eslint-config-hypothesis": "^2.6.0",
-    "eslint-plugin-mocha": "^10.4.1",
-    "eslint-plugin-react": "^7.34.1",
-    "eslint-plugin-react-hooks": "^4.6.0",
+    "eslint": "^9.12.0",
+    "eslint-config-hypothesis": "^3.0.0",
+    "eslint-plugin-mocha": "^10.5.0",
+    "eslint-plugin-react": "^7.37.1",
+    "eslint-plugin-react-hooks": "^5.0.0",
```
acelaya, line 31: This project does not really need these two plugins, but they are `eslint-config-hypothesis` peer dependencies.

robertknight: We could add separate entry points in future for Preact and non-Preact using libs/apps.

eslint.config.js
```diff
@@ -0,0 +1,83 @@
...
+      'src/chrome/content',
```
robertknight, line 17: The `src/chrome/` directory doesn't exist any longer.

eslint.config.js
```diff
@@ -0,0 +1,83 @@
...
+    },
+  },
+
+  // Scripts
+  {
+    files: [
+      'tests/karma.config.cjs',
+      'tools/*.js',
+      // Entry points which currently get loaded as non-module scripts.
+      'src/unload-client.js',
```
robertknight, line 60: `src/unload-client.js` is unlike the other scripts here because it doesn't run in Node. It turns in the browser in a web page. The main difference from other source files is that it is a standalone non-module script due to limitations of Chrome APIs.

We have these categories of files for linting:
- Source files compiled as ESM modules into a bundle
- Classic scripts that run in the browser
- CommonJS scripts that run in Node
- Module scripts that run in Node

Something like this:

```diff
diff --git a/eslint.config.js b/eslint.config.js
index 92250c81..39230686 100644
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -14,7 +14,6 @@ export default tseslint.config(
       'docs/_build/*',
       // TODO - Lint these files
       'rollup*.config.js',
-      'src/chrome/content',
     ],
   },
   ...hypothesis,
@@ -51,16 +50,25 @@ export default tseslint.config(
     },
   },
 
-  // Scripts
+  {
+    // Entry points which get loaded as non-module scripts.
+    files: ['src/unload-client.js'],
+    rules: {
+      strict: ['error', 'global'],
+    },
+    languageOptions: {
+      parserOptions: {
+        sourceType: 'script',
+      },
+    },
+  },
+
+  // CommonJS scripts which run in Node
   {
     files: [
       'tests/karma.config.cjs',
-      'tools/*.js',
-      // Entry points which currently get loaded as non-module scripts.
-      'src/unload-client.js',
     ],
     rules: {
-      'no-console': 'off',
       strict: ['error', 'global'],
     },
     languageOptions: {
@@ -73,11 +81,16 @@ export default tseslint.config(
     },
   },
 
-  // Tools
+  // ESM scripts which run in Node
   {
     files: ['tools/*.js'],
     rules: {
-      strict: 'off',
+      'no-console': 'off',
+    },
+    languageOptions: {
+      globals: {
+        ...globals.node,
+      },
     },
   },
 );
```

robertknight (changes requested):
We need to separate out `unload-client.js` from the other scripts because that doesn't run in Node. All the scripts in `tools/` _are_ ESM modules. Only `tests/karma.config.cjs` is ESM.

eslint.config.js
acelaya: Yep, makes sense

PR #1725, opened 2025-04-24
Migrate tests from karma to vitest

acelaya:
> Depends on https://github.com/hypothesis/frontend-build/pull/722
> Depends on https://github.com/hypothesis/frontend-testing/pull/76
> Depends on https://github.com/hypothesis/frontend-build/pull/736

This PR is the equivalent to https://github.com/hypothesis/frontend-shared/pull/1947, but for browser-extension repository.

### TODO

- [x] Do not dump a summary of every successful test in the output (https://github.com/hypothesis/frontend-testing/pull/76)
- [x] Fix failing tests


package.json
acelaya, line 1: I had to install autoprefixer and sass because `@hypothesis/frontend-build` has a peer dependency on them, even though we don't use them here.

I suppose I could try to run vitest similar to how we were running karma before, so that we don't need neither of those three packages, but it seemed reasonable to keep this project consistent with the rest.

robertknight: Can we make autoprefixer and sass optional peer dependencies of @hypothesis/frontend-build?

acelaya: At first I thought we couldn't without defining new entry points, but I have just realized we ship frontend-build without transpiling anything, so the "system" won't try to import either one of them unless you import the modules that use them.

So yes, we can do that. I'll prepare a patch to change that.

acelaya: Ok, no, my original assumption was correct. The lack of bundling doesn't matter, because we are still exporting everything from a common entry point, so importing from `@hypothesis/frontend-build` does end up importing all modules.

I already released a patch where autoprefixer and sass were optional, but it fails now with `Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'autoprefixer' imported from /home/alejandro/Development/hypothesis/browser-extension/node_modules/@hypothesis/frontend-build/lib/sass.js`.

I'm going to release a new version of the library with individual exports that can be used alternatively to `@hypothesis/frontend-build`, like `@hypothesis/frontend-build/test` or `@hypothesis/frontend-build/sass`, and allows to avoid the issue.

acelaya: This should fix it https://github.com/hypothesis/frontend-build/pull/740

I tested it here with yalc, and I was able to import `runTests` from `@hypothesis/frontend-build/tests` with no "module not found" errors.

robertknight (approved):

PR #1733, opened 2025-05-08
Fix reference to undefined `process` variable when loading extension

robertknight:
The `process` variable does not exist at all when the background script is loaded in Chrome normally. Resolve this in a similar way to the @hypothesis/lms repo [^1] by defining a variable using @rollup/plugin-replace instead.

Using `process.env.NODE_ENV` works in some of our other projects because they use @rollup/plugin-replace to replace it at build time. This is non-obvious from looking at the code however, so I think we should probably use a more-obvious custom variable.

[^1]: https://github.com/hypothesis/lms/blob/bc01ef9b7015b4c460250bf65b311b014e17368f/rollup-tests.config.js#L44

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "@rollup/plugin-json": "^6.1.0",
     "@rollup/plugin-multi-entry": "^6.0.1",
     "@rollup/plugin-node-resolve": "^16.0.1",
+    "@rollup/plugin-replace": "^6.0.2",
     "@types/chrome": "^0.0.319",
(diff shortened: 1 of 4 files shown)


acelaya:
Good catch. I completely missed that 🤦🏼

acelaya:
> Using process.env.NODE_ENV works in some of our other projects because they use @rollup/plugin-replace to replace it at build time. This is non-obvious from looking at the code however, so I think we should probably use a more-obvious custom variable.

Agree

PR #1750, opened 2025-05-22
Remove references to cdn.hypothes.is from boot script

robertknight:
When packaging the Hypothesis client for inclusion in the extension, replace the boot script containing cdn.hypothes.is URLs with one that references client-bundled URLs via `chrome.runtime.getURL`. We still set `assetRoot` and related URLs in order to support annotation-enabled same-origin iframes, so the new URLs are not actually used. Importantly however, the remote CDN URLs no longer appear in the bundle. This avoids giving the mistaken impression that we were loading remote code.

See https://github.com/hypothesis/client/pull/7099.

CI will fail until the client is updated to include https://github.com/hypothesis/client/pull/7099.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,9 @@ build/manifest.json: src/manifest.json.mustache build/settings.json
 build/client/build: node_modules/hypothesis/build/manifest.json
 	@mkdir -p $@
 	cp -R node_modules/hypothesis/build/* $@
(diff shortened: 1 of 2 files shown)


acelaya (approved):

acelaya:
Approved, assuming it works after https://github.com/hypothesis/client/pull/7099 has been merged

